PR #263, opened 2024-05-23
feat(wallet): add versioned path to wallet ui

keplervital:
Since station canisters that the wallet UI connects with can be at many different versions, even forward compatible changes make it difficult since the UI could expect a newer API then what it got.

This PR addresses that by introducing versioned UIs that keep track of the api version they are compatible with. It adds a compatibility layer that upon connecting to a different station canister, fetch's it's version and validates what's the best UI to use with it.

The `compatibility.core` is where most of the logic happens, however, i've also had to refactor the vite build to be able to operate seamlessly with the many versioned paths.


apps/wallet/src/core/compatibility.core.ts
```diff
@@ -0,0 +1,248 @@
...
+  const exactPathUrl = new URL(window.location.href);
+  exactPathUrl.pathname = '/' + requestedVersion + '/';
+
+  window.location.href = exactPathUrl.href;
+}
+
+/**
+ * Fetch the compatibility file for the UI, if a version is provided, then fetch the versioned file.
+ */
+async function fetchCompatFile(versionPath?: string): Promise<{
```
olaszakos, line 91: use the type in [with-compatibility-file.plugin.ts](https://github.com/dfinity/orbit/pull/263/files#diff-5c73c14e51f334fa4113a543c72bf3172a46883b2423db5c1a86d4b2e1010e04)?

apps/wallet/src/configs/init.config.ts
```diff
@@ -1,11 +1,26 @@
 import type { AppInitConfig } from '~/types/app.types';
 import { defaultLocale, supportedLocales } from './i18n.config';
+import { isSemanticVersion } from '~/utils/helper.utils';
+
+const baseUrl = import.meta.env.BASE_URL || '/';
+
+let versionedBaseUrl = baseUrl;
+const currentPath = location.pathname;
+const parsedPath = currentPath.startsWith(baseUrl)
```
olaszakos, line 9: use removeBasePathFromPathname?

olaszakos (approved):

apps/wallet/src/core/compatibility.core.ts
keplervital: makes sense, extracted and added to `build/types/compat.types.ts`

PR #264, opened 2024-05-28
fix(ci): Sanitize PR title by using ENV variable

jwndlng:
We need to sanitize the PR title by leveraging an environment variable as an intermediate step to prevent potential script injection attacks. This additional measure enhances the security of the workflow.


keplervital (approved):
LGTM

PR #270, opened 2024-05-30
feat(station): introduce ChangeManagedCanister request type

mraszyk:
This PR introduces `ChangeManagedCanister` request type with per canister resource and request specifiers.


apps/wallet/src/generated/station/station.did
```diff
@@ -1567,6 +1603,17 @@ type ChangeCanisterResourceAction = variant {
...
 
+// The target canister of a request to change a managed canister.
+type ChangeManagedCanisterResourceTarget = variant {
+  Any;
+  Canister : principal;
+};
+
+// The actions that are available for managed canister changes.
+type ChangeManagedCanisterResourceAction = variant {
+  Create : ChangeManagedCanisterResourceTarget;
```
keplervital, line 1614: What do you think of changing the approach here a bit?since we are adding a separate operation anyway and we will need to do operations like Create the canister for the first time and then also Manage it, what do you think of

```
type ManagedCanisterResourceAction = variant {
  Create;
  Change : ManagedCanisterResourceTarget;
}
```

This way we would later be able to expand this with something like `Call`, etc... and the operations for those would remain separate but mapping to the expective action, e.g.

```
ManagedCanisterResourceAction::Create -> CreateManagedCanisterOperation
ManagedCanisterResourceAction::Change -> ChangeManagedCanisterOperation
ManagedCanisterResourceAction::Call -> CallManagedCanisterOperation
```

core/station/api/spec.did
```diff
@@ -1567,6 +1603,17 @@ type ChangeCanisterResourceAction = variant {
...
 
+// The target canister of a request to change a managed canister.
+type ChangeManagedCanisterResourceTarget = variant {
+  Any;
+  Canister : principal;
+};
+
+// The actions that are available for managed canister changes.
+type ChangeManagedCanisterResourceAction = variant {
+  Create : ChangeManagedCanisterResourceTarget;
```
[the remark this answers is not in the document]
mraszyk: Creation does not combine with the Canister ID specifiers since there's no canister ID yet. Calls etc. are an option though and I could add them here. Alternatively, we could introduce a new request type for calls.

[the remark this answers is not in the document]
mraszyk: Sounds good. Let me do it tomorrow morning.

core/station/impl/src/mappers/request_policy.rs
```diff
@@ -398,6 +405,11 @@ impl RequestSpecifier {
             RequestSpecifier::ChangeCanister => vec![Resource::ChangeCanister(
                 ChangeCanisterResourceAction::Create,
             )],
+            RequestSpecifier::ChangeManagedCanister(target) => {
+                vec![Resource::ManageCanister(
+                    ManagedCanisterResourceAction::Change(target.clone()),
+                )]
+            }
```
mraszyk, line 412: @keplervital Please double check this part.

core/station/api/spec.did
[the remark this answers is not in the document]
mraszyk: Done!

core/station/api/spec.did
```diff
@@ -1567,6 +1603,23 @@ type ChangeCanisterResourceAction = variant {
...
+
+// The target canister of a request to change a managed canister.
+type ChangeManagedCanisterResourceTarget = variant {
+  Any;
+  Canister : principal;
+};
+
+// The actions that are available for managed canisters.
+type ManagedCanisterResourceAction = variant {
+  Create : CreateManagedCanisterResourceTarget;
```
keplervital, line 1619: is the plan here to extend in the future so that users would be able to set an existing canister? to, for example be able to import an existing one?

core/station/api/spec.did
```diff
@@ -1600,6 +1653,7 @@ type Resource = variant {
   Account : AccountResourceAction;
   AddressBook : ResourceAction;
   ChangeCanister : ChangeCanisterResourceAction;
+  ManageCanister : ManagedCanisterResourceAction;
```
keplervital, line 1656: Should this be like this instead?

```suggestion
  ManagedCanister : ManagedCanisterResourceAction;
```

core/station/impl/src/factories/requests/change_canister.rs
```diff
@@ -116,24 +117,81 @@ impl Execute for ChangeCanisterRequestExecute<'_, '_> {
...
+                .title
+                .unwrap_or_else(|| "ChangeManagedCanister".to_string()),
+            input.summary,
+        );
+
+        Ok(request)
+    }
+}
+
+pub struct ChangeManagedCanisterRequestExecute<'p, 'o> {
```
keplervital, line 164: ideally this would use dependency injection for the `CHANGE_CANISTER_SERVICE` here instead of the inline call

tests/integration/Cargo.toml
```diff
@@ -19,3 +19,4 @@ serde = { workspace = true, features = ['derive'] }
 control-panel-api = { path = '../../core/control-panel/api', version = '0.0.2-alpha.2' }
 upgrader-api = { path = '../../core/upgrader/api', version = '0.0.2-alpha.2' }
 station-api = { path = '../../core/station/api', version = '0.0.2-alpha.2' }
+wat = "1.0.52"
```
keplervital, line 22: ideally this should be in the workspace root Cargo.toml and then this inner Cargo.toml would just need `wat = { workspace = true }`

core/station/api/spec.did
```diff
@@ -829,6 +863,8 @@ type ListRequestsOperationType = variant {
   RemoveUserGroup;
   // An operation for changing a canister.
   ChangeCanister;
+  // An operation for changing a managed canister.
+  ChangeManagedCanister;
```
keplervital, line 867: Ideally you would be able to filter here by any or a specific canister id, similar to what we have in the transfer on line 843

core/station/api/spec.did
mraszyk: The future extension I thought about are specific subnets.

core/station/api/spec.did
mraszyk: We have ChangeCanister as another variant so we're using verbs as the variant: the variant is "manage a canister" and the specified is an "action" for this "managed canister".

core/station/impl/src/factories/requests/change_canister.rs
keplervital: i meant something like [this](https://github.com/dfinity/orbit/blob/fa4771d49c1f453325d00c5c968af18fbe930a69/core/station/impl/src/factories/requests/edit_request_policy.rs#L58)

core/station/api/spec.did
keplervital: i see your point, i was viewing the pattern more like an association, so, if not specific you would just have `ResourceAction`, but if specific you would use the same name as in the variant with the value but would suffix that name with `ResourceAction`. 

So in that case it would be `ManagedCanister` as the resource, and the actions that you can perform on that would be `ManagedCanisterResourceAction`.

The convention would be:

```
type Resource = variant {
  Example: ExampleResourceAction,
}
```

core/station/impl/src/factories/requests/change_canister.rs
mraszyk: I see. Fixed for both cases here: https://github.com/dfinity/orbit/pull/270/commits/79d308c089e1a4a3717a7805b9d1d7fb7ec7909b

core/station/api/spec.did
mraszyk: Makes sense! Fixed.

keplervital (approved):

PR #274, opened 2024-06-03
feat(control-panel): add registry schema to the control-panel

keplervital:
This PR adds the initial models and validation for registry entries and the corresponding artifacts.


core/control-panel/impl/src/models/artifact.rs
```diff
@@ -0,0 +1,107 @@
...
+pub type ArtifactId = UUID;
+
+/// The artifact is a record that is stored in the artifact repository.
+#[storable]
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Artifact {
+    /// The UUID identifies an artifact.
+    id: ArtifactId,
+    /// The artifact byte size.
+    size: u64,
```
mraszyk, line 19: Why do we store it instead of calling `artifact.len()`?

core/control-panel/impl/src/models/registry.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub namespace: String,
+    /// The name of the entry, which is used to identify it (e.g. station), within each namespace the
+    /// name should refer to a the same type of entry, but many entries can exist with the same name.
+    ///
+    /// e.g. if the namespace is "@orbit" and the name is "station", then all the entries
+    /// will refer to a wasm module that is a station.
+    pub name: String,
+    /// The tags are used to tag the entry with specific search terms (e.g. "latest", "stable").
+    ///
+    /// Tags are grouped with the format `namespace/name` (e.g. "@orbit/station").
```
mraszyk, line 37:
```suggestion
    /// Tags are grouped within the same `namespace/name` (e.g. "@orbit/station").
```

core/control-panel/impl/src/models/registry.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub tags: Vec<String>,
+    /// The category is used to associate the entry with a specific category (e.g. "chain-fusion")
+    /// across all the entries in the registry.
+    pub categories: Vec<String>,
```
mraszyk, line 41: Should these be part of metadata? (We don't seem to make any programmatic assertions about them anyway.)

core/control-panel/impl/src/models/registry.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Wether or not the entry is public, if the entry is public then it is readable by anyone.
+    pub is_public: bool,
+    /// The namespace of the entry, which is used to group entries together (e.g. "@orbit").
+    pub namespace: String,
+    /// The name of the entry, which is used to identify it (e.g. station), within each namespace the
+    /// name should refer to a the same type of entry, but many entries can exist with the same name.
+    ///
+    /// e.g. if the namespace is "@orbit" and the name is "station", then all the entries
+    /// will refer to a wasm module that is a station.
+    pub name: String,
```
mraszyk, line 34: Why do we need to distinguish a namespace and a name? What would be an example?

core/control-panel/impl/src/models/artifact.rs
keplervital: It was mostly added for clarify, but since we need to load the actual content to be able to access that maybe we could just add that to the artifact api instead of storing that as a value

core/control-panel/impl/src/models/registry.rs
keplervital: I think we would add some limits to the number of categories a registry entry can have similar to crates.io, and also, like that would be simpler to create an index with the categories.

core/control-panel/impl/src/models/registry.rs
keplervital: This is mostly helpful to search for all things included in a given namespace, this way we can tell that the orbit namespace would include the station and upgrader for example, without it we would need to either have some search by prefix, or to enforce that convention in the name itself 🤔

core/control-panel/impl/src/models/artifact.rs
mraszyk: > It was mostly added for clarify, but since we need to load the actual content to be able to access that maybe we could just add that to the artifact api instead of storing that as a value

Indeed. Unlike hashing, this should be constant time to evaluate if I'm not mistaken.

PR #275, opened 2024-06-04
feat(station): add canister_status endpoint

mraszyk:


apps/wallet/src/generated/station/station.did
```diff
@@ -1989,6 +2029,8 @@ service : (opt SystemInstall) -> {
   get_user_group : (input : GetUserGroupInput) -> (GetUserGroupResult) query;
   // List all user groups of the station.
   list_user_groups : (input : ListUserGroupsInput) -> (ListUserGroupsResult) query;
+  // Get canister status of a canister controlled by the station.
+  canister_status : (input : CanisterIdRecord) -> (CanisterStatusResult);
```
keplervital, line 2033: The current format for inputs within our spec is to use `Input` as a suffix, for this case it would mean, `CaniterStatusInput`, and even though the existing api for canister status uses `CanisterIdRecord` i would assume using `CaniterStatusInput` as a name changes nothing and should still be compatible no?

core/station/impl/src/services/status.rs
```diff
@@ -0,0 +1,34 @@
...
+use lazy_static::lazy_static;
+use orbit_essentials::api::ServiceResult;
+use std::sync::Arc;
+
+lazy_static! {
+    pub static ref STATUS_SERVICE: Arc<StatusService> = Arc::new(StatusService::default());
+}
+
+#[derive(Default, Debug)]
+pub struct StatusService {}
```
keplervital, line 14: instead of creating granular services maybe we should just create a `ManagedCanisterService` and include there the `canister_status` call and `create_canister` that we introduced in a recent PR, wdyt?

apps/wallet/src/generated/station/station.did
mraszyk: Indeed. Renaming type synonyms in the did file is harmless => done.

keplervital (approved):

PR #277, opened 2024-06-07
feat(control-panel): add registry api interface and repository

keplervital:
This PR adds the registry api interface and repositories for the registry. I've only included some of the get implementation in the service layer for now and had to move the pagination from the station canister to the common orbit_essentials lib to be reused by the control-panel. 

Also, i'll still create a follow up PR that will remove the pagination implementation from the station canister and will start to use it from the common orbit_essentials lib.

### DIFF
--- a/core/control-panel/api/spec.did
+++ b/core/control-panel/api/spec.did
@@ -383,7 +383,7 @@ type RegistryEntry = record {
   // The timestamp when the entry was created.
   created_at : TimestampRFC3339;
   // The timestamp when the entry was last updated.
-  updated_at : TimestampRFC3339;
+  updated_at : opt TimestampRFC3339;
 };
 
 // The registry entry value.
@@ -408,15 +408,15 @@ type WasmModuleRegistryEntryValue = record {
 };
 
 // The search registry filter options.
-type SearchRegistryFilter = variant {
+type SearchRegistryFilterKind = variant {
   // The name of the registry entry to find, if the namespace is not provided the default namespace is used.
-  name : text;
+  Name : text;
 };
 
 // The input for searching the registry.
 type SearchRegistryInput = record {
-  // Filters are used sequentially to filter the entries in the registry.
-  filter_by : vec SearchRegistryFilter;
+  // Filters are used sequentially and are combined with an AND operation to filter the entries in the registry.
+  filter_by : vec SearchRegistryFilterKind;
   // The pagination options to use for the search.
   pagination : opt PaginationInput;
 };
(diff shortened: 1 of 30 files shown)


core/control-panel/api/spec.did
```diff
@@ -310,8 +319,254 @@ type GetArtifactResult = variant {
...
 };
 
+// A metadata record that contains a key and a value.
+type Metadata = record {
+  key : text;
+  value : text;
+};
+
+// An entry record, which contains information and the value stored in the registry.
+type RegistryEntry = record {
```
olaszakos, line 329: These were renamed in Rust, intended to for Candid?

core/control-panel/api/spec.did
```diff
@@ -310,8 +319,254 @@ type GetArtifactResult = variant {
...
+  //
+  // e.g. if the namespace is "@orbit" and the name is "station", then all the entries will refer to a wasm module.
+  //
+  // Restrictions:
+  //
+  // - Names that start with `@` are considered namespaced.
+  // - Names that start with `@` must have a namespace and a name separated by a `/`.
+  // - Names must be between 2 and 48 characters long.
+  // - Namespaces must be between 2 and 32 characters long.
+  // - Names that are not namespaced, are put in the default namespace `@a`.
```
olaszakos, line 344: default namespace is `@default`, same for the comment in the Rust counterpart

core/control-panel/impl/src/models/registry.rs
```diff
@@ -12,21 +12,18 @@ use std::collections::BTreeMap;
...
-/// The registry entry is a record that is stored in the registry repository.
+/// The registry is a record that is stored in the registry repository.
 ///
 /// It stores entries about wasm modules and can be extended to other entry types. When adding new entry types,
-/// the `RegistryEntryValue` enum should be updated to include the new entry type.
+/// the `RegistryValue` enum should be updated to include the new entry type.
 #[storable]
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
-pub struct RegistryEntry {
+pub struct Registry {
```
olaszakos, line 24: To me Registry sounds like a collection of items, not the item itself in the collection.

olaszakos (approved):

core/control-panel/impl/src/models/registry.rs
keplervital: Makes sense, i'll rename this back to `RegistryEntry`

core/control-panel/api/spec.did
keplervital: good catch

PR #281, opened 2024-06-10
feat(station): add CallExternalCanister request type

mraszyk:
This PR adds a `CallExternalCanister` request type.


core/station/api/spec.did
```diff
@@ -592,6 +593,54 @@ type CreateExternalCanisterOperation = record {
...
+  // }
+  // ```
+  // If omitted (`validation_method = null`), no validation of the argument blob is performed
+  // and no human-readable rendering of the argument blob is provided.
+  validation_method : opt CanisterMethod;
+  // The canister method that is called after the request becomes `Approved`
+  // passing the validated argument blob.
+  execution_method : CanisterMethod;
+  // The argument blob passed to both the validation and execution method.
+  arg : blob;
```
keplervital, line 623: should the argument be optional?

mraszyk: Do you think it's gonna be common to call endpoints without any argument?

keplervital: only if you have triggering endpoints, e.g. `unlock`, `lock`, etc...

mraszyk: Sure, I've made the argument blob optional now.

core/station/impl/src/core/validation.rs
```diff
@@ -188,7 +192,7 @@ impl EnsureExternalCanister {
         principal: Principal,
     ) -> Result<(), ExternalCanisterValidationError> {
         if principal == Principal::management_canister()
-            || principal == ic_cdk::id()
+            || principal == ic_cdk::api::id()
             || principal == SYSTEM_SERVICE.get_upgrader_canister_id()
```
keplervital, line 196: we should also add here the ICP ledger which is the only ledger we support currently

mraszyk: Done: https://github.com/dfinity/orbit/pull/281/commits/1a5126cc7ab81d6cb7a6d1cbdf4ed60208d66444

keplervital (approved):

bitdivine:
Does this mean that I can do asset canister upgrades now?

keplervital:
> Does this mean that I can do asset canister upgrades now?

Yes, indeed!

PR #283, opened 2024-06-17
feat(control-panel): add registry api implementation

keplervital:
This PR adds the implementation for the registry api, enabling admins of the control panel to add versioned wasm modules that consumers of the API can then fetch through the artifact api.


core/control-panel/impl/src/models/registry_entry.rs
```diff
@@ -308,24 +418,14 @@ fn validate_name(name: &str) -> ModelValidatorResult<RegistryError> {
...
-    }
-
-    if (name.len() < RegistryEntry::MIN_NAME_LENGTH)
-        || (name.len() > RegistryEntry::MAX_NAME_LENGTH)
-    {
-        return Err(RegistryError::ValidationError {
-            info: format!(
-                "Name length must be between {} and {}",
                 RegistryEntry::MIN_NAME_LENGTH,
                 RegistryEntry::MAX_NAME_LENGTH,
```
olaszakos, line 422:
```suggestion
                RegistryEntry::MIN_NAMESPACE_LENGTH,
                RegistryEntry::MAX_NAMESPACE_LENGTH,
```

core/control-panel/impl/src/models/registry_entry.rs
```diff
@@ -519,25 +625,47 @@ pub mod registry_entry_test_utils {
...
     use registry_entry_test_utils::create_registry_entry;
     use rstest::rstest;
     use uuid::Uuid;
 
+    #[rstest]
+    #[case::empty_namespace(&"")]
+    #[case::empty_namespace_with_space(&" ")]
+    #[case::starts_with_space(&" orbit")]
+    #[case::ends_with_space(&"orbit ")]
+    #[case::namespace_to_small(&"a")]
```
olaszakos, line 638:
```suggestion
    #[case::namespace_too_small(&"a")]
```

core/control-panel/impl/src/models/registry_entry.rs
```diff
@@ -519,25 +625,47 @@ pub mod registry_entry_test_utils {
...
+
+        validate_namespace(&entry.namespace).unwrap();
+    }
+
     #[rstest]
     #[case::empty_name(&"")]
     #[case::empty_name_with_space(&" ")]
     #[case::starts_with_space(&" entry")]
     #[case::ends_with_space(&"entry ")]
     #[case::name_to_small(&"a")]
```
olaszakos, line 667:
```suggestion
    #[case::name_too_small(&"a")]
```

core/control-panel/impl/src/repositories/registry.rs
```diff
@@ -75,169 +89,262 @@ impl Repository<RegistryEntryId, RegistryEntry> for RegistryRepository {
...
+
+        id_with_timestamps.sort_by(|a, b| {
+            {
+                let ord = a.0.cmp(&b.0); // Compare timestamps
+                match direction {
+                    SortDirection::Ascending => ord,
+                    SortDirection::Descending => ord.reverse(),
+                }
+            }
+            .then_with(|| a.1.cmp(&b.1)) // Compare IDs if timestamps are equal
```
olaszakos, line 260: reverse direction for Descending?

keplervital: makes sense, just pushed the change

core/control-panel/impl/src/services/registry.rs
```diff
@@ -48,14 +58,146 @@ impl RegistryService {
...
+            previous_latest
+                .iter()
+                .filter_map(|id| self.get(id).ok())
+                .for_each(|mut e| {
+                    e.tags.retain(|t| t != LATEST_TAG);
+                    self.registry_repository.insert(e.id, e);
+                });
+        }
+
+        entry.validate()?;
```
olaszakos, line 156: remove artifact if validation failes?

core/control-panel/impl/src/services/registry.rs
```diff
@@ -48,14 +58,146 @@ impl RegistryService {
...
+                        .dependencies
+                        .iter()
+                        .map(|dep| dep.clone().into())
+                        .collect(),
+                });
+            }
+            (None, _) => (),
+        };
+
+        entry.validate()?;
```
olaszakos, line 196: if this fails then the original entry will stay but its artifact is already deleted, and the new one is dangling.

core/control-panel/impl/src/services/registry.rs
olaszakos: also, the `latest` tag could be lost if validation fails.

core/control-panel/impl/src/services/registry.rs
olaszakos: does `latest` tag need check here? maybe it's added by this edit.

olaszakos (approved):

core/control-panel/impl/src/services/registry.rs
keplervital: Great point, since we're not adding a `panic` there, just made a change that accounts for that

core/control-panel/impl/src/services/registry.rs
keplervital: Great point, just made a change that accounts for that

PR #284, opened 2024-06-18
feat(api): Add the serde::Serialize trait to API types

bitdivine:
# Motivation
The types defined in the API are for external use.  These types would be considerably more useful if they were to implement `serde` serialize and deserialize traits.  An example use is the equivalent of `dfx canister call --output json` in the upcoming Orbit command line tool.

The `serde::Deserialize` trait is already derived for use with candid.  Only the `serde::Serialize` trait needs to be added.

Possible counterarguments:

- The Wasm sizes may increase.
  - I tested the station canister, as that has the largest number of changed types.  The Wasm size is unchanged, even the sha256sum of the compiled Wasm is unchanged:
    - Current main:

          max@sinkpad:~/dfn/orbit (2:21)$ ls -lh wasms/station.wasm.gz 
          -rw-rw-r-- 1 max max 1,5M Jun 18 11:20 wasms/station.wasm.gz

          max@sinkpad:~/dfn/orbit (2:27)$ ls -l wasms/station.wasm.gz 
          -rw-rw-r-- 1 max max 1549782 Jun 18 11:20 wasms/station.wasm.gz

          max@sinkpad:~/dfn/orbit (5:27)$ sha256sum wasms/station.wasm.gz
          25b027994ae7c0c267cd8cae442b5571820c25bc5dfc9d15fa16783204c45782  wasms/station.wasm.gz
    - This PR:

          max@sinkpad:~/dfn/orbit/branches/serde (6:56)$ ls -lh wasms/station.wasm.gz
          -rw-rw-r-- 1 max max 1,5M Jun 18 11:24 wasms/station.wasm.gz

          max@sinkpad:~/dfn/orbit/branches/serde (6:59)$ ls -l wasms/station.wasm.gz
          -rw-rw-r-- 1 max max 1549782 Jun 18 11:24 wasms/station.wasm.gz

          max@sinkpad:~/dfn/orbit/branches/serde (7:38)$ sha256sum wasms/station.wasm.gz
          25b027994ae7c0c267cd8cae442b5571820c25bc5dfc9d15fa16783204c45782  wasms/station.wasm.gz

- Third parties will always want some trait.  We cannot cater for them all.  If they want to add a trait, let them clone the type and add the trait.
  - True, but here Orbit is not adding any third party libraries not already used by Orbit.  Likewise third parties using this API already need serde for the candid trait.  No `Cargo.lock` will need to be updated as a result of this PR (unless we bump the semantic version).  Accommodating serde usage is not an imposition either on Orbit or users of the Orbit API.

# Changes
- Add the `serde::Serialize` trait to every `CandidType` in `core/*/api/`
- Formatting: `cargo sort` sorted the root `Cargo.toml`

# Tests
By the fact that the code compiles I believe that we can be confident that `serde::Serialize` is indeed available on these types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,17 +1,17 @@
 [workspace]
 resolver = "2"
 members = [
-    "tests/integration",
     "core/control-panel/api",
     "core/control-panel/impl",
-    "core/upgrader/api",
-    "core/upgrader/impl",
     "core/station/api",
     "core/station/impl",
+    "core/upgrader/api",
+    "core/upgrader/impl",
     "libs/canfund",
     "libs/orbit-essentials",
     "libs/orbit-essentials-macros",
     "libs/orbit-essentials-macros-tests",
+    "tests/integration",
 ]
 
 [workspace.package]
--- a/core/control-panel/api/src/artifact.rs
+++ b/core/control-panel/api/src/artifact.rs
@@ -1,12 +1,12 @@
 use crate::{Sha256HashDTO, TimestampRfc3339, UuidDTO};
 use candid::{CandidType, Deserialize};
 
-#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[derive(CandidType, serde::Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub struct GetArtifactInput {
     pub artifact_id: UuidDTO,
 }
 
-#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[derive(CandidType, serde::Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub struct ArtifactDTO {
     pub id: UuidDTO,
     pub size: u64,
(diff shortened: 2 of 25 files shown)


Cargo.toml
```diff
@@ -1,17 +1,17 @@
...
-    "core/upgrader/impl",
     "core/station/api",
     "core/station/impl",
+    "core/upgrader/api",
+    "core/upgrader/impl",
     "libs/canfund",
     "libs/orbit-essentials",
     "libs/orbit-essentials-macros",
     "libs/orbit-essentials-macros-tests",
+    "tests/integration",
```
bitdivine, line 14: Formatter at work?  I probably have `cargo sort` in my default pre-commit hook.  I can un-sort these if wished. :-)

core/control-panel/api/src/artifact.rs
```diff
@@ -1,12 +1,12 @@
 use crate::{Sha256HashDTO, TimestampRfc3339, UuidDTO};
 use candid::{CandidType, Deserialize};
 
-#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[derive(CandidType, serde::Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
```
bitdivine, line 4: I programatically added `serde::Serialize` to every `derive(CandidType` and `, CandidType` in the `core/*/api` directories.

Cargo.toml
keplervital: No need, that's fine

keplervital (approved):
LGTM

**fyi** i've added the scope `api` to the PR title, the tool that we use (nx) relies on this to not bump all the semantic versions in all packages in the repo, this way it only bumps them if the touched files affect the package

PR #286, opened 2024-06-21
feat(station): optional fallback controller

mraszyk:
This PR adds an optional fallback controller to the station's initial arguments (used when installing/reinstalling the station).  The control panel sets the NNS root canister as the fallback controller.


keplervital (approved):
LGTM

PR #289, opened 2024-07-02
feat(cli): helper cli commands to interact with the control-panel registry

keplervital:
This PR adds some helper commands to the orbit dev cli, facilitating the interaction with the control panel registry for publishing the upgrader and station canisters.

Commands added under `orbit-cli registry ...`:

- `list`: Lists all the versions of the app provided
- `remove`: Remove a single entry in the registry
- `publish`: Publishes the provided app (upgrader/station) to the registry, in the case of the station the upgrader is also published since it is a dependency


cli/src/registry/publish.ts
```diff
@@ -0,0 +1,227 @@
...
+};
+
+command.action(async options => {
+  const ping: { replica_health_status?: string } = JSON.parse(
+    await execAsync(`dfx ping '${options.network}'`),
+  );
+
+  if (ping.replica_health_status?.toLowerCase() !== 'healthy') {
+    throw new Error('The replica is not healthy.');
+  }
```
nathanosdev, line 120: Why not use `assertReplicaIsHealthy` here?

cli/src/registry/publish.ts
```diff
@@ -0,0 +1,227 @@
...
+  description: string;
+  tags: string[];
+  categories: string[];
+  metadata: Record<string, string>;
+  version: string;
+  module: Uint8Array;
+  dependencies: PackageDedependency[];
+}
+
+const getAppRegistryEntry = async (app: Application, network: string): Promise<Package> => {
```
nathanosdev, line 35: There's no default cause in the switch statement so the correct return type should be `Promise<Package | undefined>`. Maybe a simple `throw` would be nice to help debug it in case it somehow goes wrong. I think it's fine though since it's just a CLI for local use.

cli/src/registry/publish.ts
```diff
@@ -0,0 +1,227 @@
...
+  const upgraderVersion = cargoProjectVersion(Application.Upgrader);
+  console.log(`Creating release artifacts for project: upgrader...`);
+  execSync(`BUILD_MODE='${network}' npx nx run upgrader:create-artifacts`);
```
nathanosdev, line 38: If the selected app is the station then these artifacts will be generated twice, once when it's called for the station, and again when it's called for the upgrader as a dependency of the station. Is that intended?

nathanosdev (approved):
LGTM, all comments are optional for me.

cli/src/registry/publish.ts
keplervital: Missed that one, just added now

cli/src/registry/publish.ts
keplervital: Typescript is still happy because `Application` has only 2 variants and both are covered in the switch, but just to be on the safe side i added the default with a throw as you suggested.

cli/src/registry/publish.ts
keplervital: Initially yes, but now that i added the dependencies there is no need for that anymore, nice catch. I moved it to the case condition.

cli/package.json
```diff
@@ -8,9 +8,14 @@
   "scripts": {
     "build": "tsc",
     "start": "node dist/cli.js",
-    "expose": "pnpm link --global"
+    "expose": "pnpm link --global",
+    "generate-types": "didc bind --target ts ../core/control-panel/api/spec.did > ./src/generated/control_panel.d.ts"
```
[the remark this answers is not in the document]
keplervital: iirc `dfx generate` uses the `declarations` output path defined in `dfx.json`, and we have that set for the wallet application. Also, dfx generate also adds the actors, and we simply need the types from the spec here.

cli/src/generated/control_panel.d.ts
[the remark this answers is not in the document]
keplervital: This is mainly because some contributors might not have `didc` available, so this unblocks such cases.

PR #290, opened 2024-07-03
build(ci): add deterministic builds

keplervital:
This PR introduces deterministic builds to the relevant canisters.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,43 +11,50 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
 # Code specific dependencies
 FROM base as builder
 WORKDIR /code
-COPY . .
 ENV RUSTUP_HOME=/opt/rustup \
     CARGO_HOME=/opt/cargo \
-    FNM_DIR=/opt/fnm \
-    MALLOC_CONF="narenas:1,abort:false"
+    FNM_DIR=/opt/fnm
 # Path modifications need to be done in separate ENV statements
 ENV PATH=$CARGO_HOME/bin:$PATH
 ENV PATH=$FNM_DIR/bin:$PATH
 # Install Rust and the Node.js version manager
+COPY rust-toolchain.toml .
+COPY .nvmrc .
 RUN curl -fsSL https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path && \
     curl -fsSL https://fnm.vercel.app/install | bash -s -- --install-dir $FNM_DIR/bin --skip-shell
+# Add the fnm env var envaluation to the bashrc to enable it in bash by default
+RUN echo "eval \"$(fnm env)\"" >> $HOME/.bashrc
 # Install the correct version of Node.js and the dependencies
+COPY . .
 RUN eval "$(fnm env)" && \
     fnm install && \
     fnm use && \
     corepack enable && \
     pnpm install
-# Add the fnm env var envaluation to the bashrc to enable it in bash by default
-RUN echo "eval \"$(fnm env)\"" >> $HOME/.bashrc
-
-# Build the Orbit Station Canister
-FROM builder as build_station
-WORKDIR /code
-LABEL io.icp.artifactType="canister" \
-      io.icp.artifactName="station"
-RUN BUILD_MODE=production npx nx run station:create-artifacts
 
 # Build the Orbit Upgrader Canister
 FROM builder as build_upgrader
 WORKDIR /code
 LABEL io.icp.artifactType="canister" \
       io.icp.artifactName="upgrader"
-RUN BUILD_MODE=production npx nx run upgrader:create-artifacts
+RUN eval "$(fnm env)" && \
+    fnm use && \
+    BUILD_MODE=production npx nx run upgrader:create-artifacts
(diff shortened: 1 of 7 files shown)


mraszyk:
Could we also add a CI job for the Docker build?

Dockerfile
```diff
@@ -0,0 +1,60 @@
+# Operating system with basic tools
+FROM --platform=linux/amd64 ubuntu:20.04 as base
```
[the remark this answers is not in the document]
keplervital: Just updated, also enforced our `Dockerfile` with the hash of the image.

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+# Install Rust and the Node.js version manager
+COPY rust-toolchain.toml .
+COPY .nvmrc .
+RUN curl -fsSL https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path && \
+    curl -fsSL https://fnm.vercel.app/install | bash -s -- --install-dir $FNM_DIR/bin --skip-shell
+# Add the fnm env var envaluation to the bashrc to enable it in bash by default
+RUN echo "eval \"$(fnm env)\"" >> $HOME/.bashrc
+# Install the correct version of Node.js and the dependencies
+COPY . .
+RUN eval "$(fnm env)" && \
```
nathanosdev, line 29: IIRC this line will run again if Docker detects any changes in any of the previous lines, and since `COPY . .` copies all files, any changes to the source files will result in all of this stuff being run again.

So maybe it would be more cache effecient to split it.

```
COPY .nvmrc .
RUN eval "$(fnm env)" && \
    fnm install && \
    fnm use

COPY package.json .
RUN corepack enable && \
    pnpm install

COPY . .
```

keplervital:
> Could we also add a CI job for the Docker build?

I would try and avoid having this as a recurring CI job for now until we add some caching to the images, otherwise the CI will take a longer then necessary and the underlying bash script that builds the wasms is already being used.

One follow up PR that is required though is to change the dev cli used [here](https://github.com/dfinity/orbit/blob/c384e83cc03b8b5db88671adaa0d16ca30585d56/.github/workflows/release.yaml#L30) to now start using the docker build artifacts when publishing a new release.

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    fnm install && \
+    fnm use && \
+    corepack enable && \
+    pnpm install
```
nathanosdev, line 33: I like using [--frozen-lockfile](https://pnpm.io/cli/install#--frozen-lockfile) to help catch issues with the lockfile being out of date early. It works like this by default in the CI environment.

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+# Build the Orbit Upgrader Canister
+FROM builder as build_upgrader
+WORKDIR /code
+LABEL io.icp.artifactType="canister" \
+      io.icp.artifactName="upgrader"
+RUN eval "$(fnm env)" && \
+    fnm use && \
+    BUILD_MODE=production npx nx run upgrader:create-artifacts
```
nathanosdev, line 42: Since I'm not familiar with the codebase I'm not sure if it's the case, but if there's shared dependencies then `nx` would be more effecient building everything in one command.

scripts/docker-build.sh
```diff
@@ -0,0 +1,124 @@
...
+
+function build_canister() {
+  local canister_name=$1
+  local target=$2
+
+  # Build the canister
+  docker build -t orbit-$canister_name --target $target .
+
+  # Create a container to extract the generated artifacts
+  docker create --name orbit-$canister_name-container orbit-$canister_name
```
nathanosdev, line 57: Is there any difference between these next lines and running it like this?

```
docker run --rm -v ./artifacts:/code/artifacts/$canister_name orbit-$canister_name-container
```

nathanosdev (approved):

Dockerfile
keplervital: Makes sense

scripts/docker-build.sh
mraszyk: With `-v` it is less clear that the Docker container doesn't just take your existing artifacts instead of building them from scratch.

Dockerfile
mraszyk: > COPY . . copies all files

there's already a .dockerignore but still more files are copied than needed

mraszyk (approved):

Dockerfile
keplervital: I made a small adjustment that improves a bit the caching of the layers, but pnpm install needs to happen after the `COPY . .` because it needs the package json from the inner monorepo projects

Dockerfile
keplervital: That's true, but this artifacts are custom commands that are mostly using cargo for wasm generation and those are not being cached yet.

scripts/docker-build.sh
keplervital: `docker cp` avoids potential file permission issues, that's the main reason i went for it.

Dockerfile
keplervital: the js environment also needs some source code in a postinstall hook that `pnpm install` triggers

PR #292, opened 2024-07-04
feat(wallet): Always lowercase ICP Ledger addresses

keplervital:


apps/wallet/src/utils/app.utils.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+} from '~/utils/app.utils';
+
+describe('BlockchainAddress', () => {
+  it('ICP Ledger account identifiers are case insensitive', () => {
+    expect(
+      isCaseInsensitiveBlockchainAddress(
+        BlockchainType.InternetComputer,
+        BlockchainStandard.Native,
+      ),
+    ).toBeTruthy();
```
nathanosdev, line 15: I think `toBe(true)` would be more appropriate here since `toBeTruthy()` will succeed for all sorts of things.

Some tests for the negative case in this and the other test here would be nice too, but I don't think it's a big deal since the logic is simple.

nathanosdev (approved):

apps/wallet/src/utils/app.utils.spec.ts
keplervital: Updated to `toBe(true)`. Since the ICP standard is the only possible option so far there's no negative case, we can add those once we support other types of addresses.

PR #293, opened 2024-07-04
feat(upgrader): disaster recovery flow reuses the same upgrader canister

keplervital:
Enables the disaster recovery flow to reuse the same upgrader canister in case the mode of recovery would be set to `install` or `reinstall`.

### DIFF
--- a/core/upgrader/impl/src/services/install_canister.rs
+++ b/core/upgrader/impl/src/services/install_canister.rs
@@ -3,7 +3,7 @@ use async_trait::async_trait;
 use candid::Principal;
 use ic_cdk::api::management_canister::main::{self as mgmt, CanisterIdRecord, InstallCodeArgument};
 use lazy_static::lazy_static;
-use std::sync::Arc;
+use std::{collections::BTreeSet, sync::Arc};
 
 lazy_static! {
     pub static ref INSTALL_CANISTER: Arc<StationDisasterRecoveryInstall> =
@@ -40,10 +40,29 @@ impl InstallCanister for StationDisasterRecoveryInstall {
         // For install and reinstall, we need to make the station self controlled.
         match mode {
             InstallMode::Install | InstallMode::Reinstall => {
+                // Get the current controllers and add the station to it. We preserve the existing controllers
+                // during this step to avoid losing control in case of a failed install.
+                let (info,) = mgmt::canister_info(mgmt::CanisterInfoRequest {
+                    canister_id,
+                    num_requested_changes: None,
+                })
+                .await
+                .map_err(|(code, err)| {
+                    format!(
+                        "failed to get canister info for canister: \"{}\", rejection code: {}",
+                        err, code as i32
+                    )
+                })?;
+
+                let mut controllers = BTreeSet::new();
+                controllers.extend(info.controllers.iter().cloned());
+                controllers.insert(canister_id);
+                controllers.insert(ic_cdk::id());
+
                 mgmt::update_settings(mgmt::UpdateSettingsArgument {
                     canister_id,
                     settings: mgmt::CanisterSettings {
-                        controllers: Some(vec![canister_id, ic_cdk::id()]),
+                        controllers: Some(controllers.into_iter().collect()),
                         ..Default::default()
                     },
                 })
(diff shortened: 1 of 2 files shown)


core/station/impl/src/services/system.rs
```diff
@@ -413,8 +411,32 @@ mod install_canister_handlers {
...
+            station_api::SystemUpgraderInput::Id(upgrader_id) => {
+                mgmt::update_settings(mgmt::UpdateSettingsArgument {
+                    canister_id: upgrader_id,
+                    settings: mgmt::CanisterSettings {
+                        controllers: Some(controllers),
+                        ..Default::default()
+                    },
+                })
+                .await
+                .map_err(|e| format!("Failed to set upgrader controller: {:?}", e))?;
```
mraszyk, line 428: I guess the point of this is to add a fallback controller if a new one is specified, right?

core/upgrader/impl/src/services/install_canister.rs
```diff
@@ -39,6 +37,29 @@ impl InstallCanister for StationDisasterRecoveryInstall {
         arg: Vec<u8>,
         mode: InstallMode,
     ) -> Result<(), String> {
+        // For install and reinstall, we need to make the station self controlled.
+        match mode {
+            InstallMode::Install | InstallMode::Reinstall => {
+                mgmt::update_settings(mgmt::UpdateSettingsArgument {
+                    canister_id: canister_id.clone(),
+                    settings: mgmt::CanisterSettings {
+                        controllers: Some(vec![canister_id, ic_cdk::id()]),
```
mraszyk, line 46: This would wipe the fallback controller if there's one.

tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -422,6 +418,85 @@ fn test_disaster_recovery_flow() {
...
+        (upgrader_api::RequestDisasterRecoveryInput {
+            module: station_wasm_module.clone(),
+            arg: Encode!(&station_api::SystemInstall::Init(station_api::SystemInit {
+                name: "Station".to_string(),
+                admins: vec![station_api::AdminInitInput {
+                    identity: WALLET_ADMIN_USER.clone(),
+                    name: "updated-admin-name".to_string(),
+                }],
+                quorum: None,
+                fallback_controller: None,
```
mraszyk, line 449: Maybe we could provide one here to check against the bug mentioned above.

core/upgrader/impl/src/services/install_canister.rs
keplervital: The station canister would do that anyway in the timer that the init hook creates [here](https://github.com/dfinity/orbit/blob/9497822f35d9361decff83d2afb34af97cc77e57/core/station/impl/src/services/system.rs#L171)

tests/integration/src/disaster_recovery_tests.rs
keplervital: The existing behaviour is to clear up the controllers in the init hook and make sure that only if a `fallback_controller` is specified that it would be set, so the test would be adding it where you suggested and then test that it is not there anymore, wdyt?

core/upgrader/impl/src/services/install_canister.rs
mraszyk: I see. So it isn't necessarily a bug after all, but if the call you linked fails, then the fallback controller is wiped for good, so it might make sense to preserve it.

keplervital: That makes sense, i'll add it here and will also extend the test mentioned below

tests/integration/src/disaster_recovery_tests.rs
mraszyk: > test that it is not there anymore

If there's a new fallback controller specified, then the station (and the upgrader) should have that new fallback controller eventually. The old fallback controller (if any) should ideally be preserved until the new fallback controller is set (the controller change should happen atomically).

core/upgrader/impl/src/services/install_canister.rs
mraszyk: The upgrader can either remember the fallback controller in its state or query it using canister_status.

mraszyk (approved):

PR #294, opened 2024-07-04
feat(upgrader): disaster recovery recreates accounts

keplervital:
This PR adds the optional functionality of providing an initial set of accounts to the init args of the station canister. This feature enables the disaster recovery to restore accounts in case of `(re)install`.

### DIFF
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -1994,7 +1994,7 @@ type SystemUpgraderInput = variant {
   WasmModule : blob;
 };
 
-// The initial accounts to create when initializing the canister for the first time.
+// The initial accounts to create when initializing the canister for the first time, e.g., after disaster recovery.
 type InitAccountInput = record {
   // The UUID of the account, if not provided a new UUID will be generated.
   id : opt UUID;
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -1994,6 +1994,20 @@ type SystemUpgraderInput = variant {
   WasmModule : blob;
 };
 
+// The initial accounts to create when initializing the canister for the first time.
+type InitAccountInput = record {
+  // The UUID of the account, if not provided a new UUID will be generated.
+  id : opt UUID;
+  // A friendly name for the account (e.g. "My Account").
+  name : text;
+  // The blockchain identifier (e.g., `ethereum`, `bitcoin`, `icp`, etc.)
+  blockchain : text;
+  // The asset standard for this account (e.g. `native`, `erc20`, etc.).
+  standard : text;
+  // Metadata associated with the account (e.g. `{"contract": "0x1234", "symbol": "ANY"}`).
+  metadata : vec AccountMetadata;
+};
+
 // The init configuration for the canister.
 //
 // Only used when installing the canister for the first time.
@@ -2008,6 +2022,8 @@ type SystemInit = record {
   upgrader : SystemUpgraderInput;
   // An optional additional controller of the station and upgrader canisters.
   fallback_controller : opt principal;
+  // Optional initial accounts to create.
+  accounts : opt vec InitAccountInput;
 };
 
 // The upgrade configuration for the canister.
--- a/core/station/impl/src/errors/system.rs
+++ b/core/station/impl/src/errors/system.rs
@@ -10,8 +10,8 @@ pub enum SystemError {
     InitFailed { reason: String },
     #[error(r#"The canister needs at least one admin"#)]
     NoAdminsSpecified,
-    #[error(r#"There are too many admins defined, max is 255."#)]
-    TooManyAdminsSpecified,
+    #[error(r#"There are too many admins defined, max allowed is {max}."#)]
+    TooManyAdminsSpecified { max: usize },
 }
 
 impl DetailableError for SystemError {
--- a/core/station/impl/src/errors/system.rs
+++ b/core/station/impl/src/errors/system.rs
@@ -10,6 +10,8 @@ pub enum SystemError {
     InitFailed { reason: String },
     #[error(r#"The canister needs at least one admin"#)]
     NoAdminsSpecified,
+    #[error(r#"There are too many admins defined, max is 255."#)]
+    TooManyAdminsSpecified,
 }
 
 impl DetailableError for SystemError {
--- a/core/station/impl/src/services/account.rs
+++ b/core/station/impl/src/services/account.rs
@@ -610,7 +610,7 @@ mod tests {
         ctx.repository.insert(account.to_key(), account.clone());
 
         let input = AddAccountOperationInput {
-            name: "foo".to_string(),
+            name: "foo2".to_string(),
             blockchain: Blockchain::InternetComputer,
             standard: BlockchainStandard::Native,
             metadata: Metadata::default(),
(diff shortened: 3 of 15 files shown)


tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -418,6 +419,189 @@ fn test_disaster_recovery_flow() {
...
+        if let RequestOperationDTO::AddAccount(operation) = request.operation {
+            let newly_added_account = operation
+                .account
+                .expect("Unexpected new account not available");
+
+            initial_accounts.insert(
+                newly_added_account.id,
+                (newly_added_account.name, newly_added_account.address),
+            );
+        }
```
mraszyk, line 480:
```suggestion
        } else { panic!(...) }
```

tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -418,6 +419,189 @@ fn test_disaster_recovery_flow() {
...
+    }
+
+    // 3. perform a reinstall disaster recovery request
+    let init_accounts_input = initial_accounts
+        .clone()
+        .into_iter()
```
mraszyk, line 486:
```suggestion
        .iter()
```

tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -418,6 +419,189 @@ fn test_disaster_recovery_flow() {
...
+            configs_permission: AllowDTO {
+                auth_scope: station_api::AuthScopeDTO::Restricted,
+                user_groups: vec![],
+                users: vec![admin_user.id.clone()],
+            },
+            transfer_permission: AllowDTO {
+                auth_scope: station_api::AuthScopeDTO::Restricted,
+                user_groups: vec![],
+                users: vec![admin_user.id.clone()],
+            },
```
mraszyk, line 457: Could you please also assert how these look like after the recovery?

tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -418,6 +419,189 @@ fn test_disaster_recovery_flow() {
...
+            transfer_request_policy: Some(RequestPolicyRuleDTO::AutoApproved),
+            configs_request_policy: Some(RequestPolicyRuleDTO::AutoApproved),
+            metadata: vec![],
```
mraszyk, line 460: Maybe we could also make sure that the metadata are recovered properly (the ones set here discarded and replaced by those in the disaster recovery input).

core/station/api/spec.did
```diff
@@ -1994,6 +1994,20 @@ type SystemUpgraderInput = variant {
   WasmModule : blob;
 };
 
+// The initial accounts to create when initializing the canister for the first time.
```
mraszyk, line 1997:
```suggestion
// The initial accounts to create when initializing the canister for the first time, e.g., after disaster recovery.
```

core/station/impl/src/services/system.rs
```diff
@@ -356,30 +361,43 @@ mod install_canister_handlers {
...
     /// Registers the default configurations for the canister.
     pub async fn init_post_process(init: &SystemInit) -> Result<(), String> {
-        let admin_count = u16::try_from(init.admins.len()).unwrap_or(u16::MAX);
-
-        let admin_quorum = init
-            .quorum
-            .unwrap_or(admin_count / 2 + 1)
-            .clamp(1, admin_count);
+        let admin_quorum = self::calc_initial_quorum(
+            u16::try_from(init.admins.len()).unwrap_or(u16::MAX),
```
mraszyk, line 398: Shouldn't we fail if there are more than u16::MAX admins rather than silently capping at u16::MAX?

core/station/impl/src/services/account.rs
```diff
@@ -522,55 +534,98 @@ mod tests {
...
+            read_permission: Allow::users(vec![ctx.caller_user.id]),
+            configs_permission: Allow::users(vec![ctx.caller_user.id]),
+            transfer_permission: Allow::users(vec![ctx.caller_user.id]),
+            configs_request_policy: Some(RequestPolicyRule::AutoApproved),
+            transfer_request_policy: Some(RequestPolicyRule::AutoApproved),
+        };
+
+        let result = ctx.service.create_account(input, Some(account.id)).await;
+
+        assert!(result.is_err());
```
mraszyk, line 626: Could we check for the error "Account ... already exists"?

tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -418,6 +419,190 @@ fn test_disaster_recovery_flow() {
...
+    let res: (ApiResult<()>,) = update_candid_as(
+        &env,
+        upgrader_id,
+        WALLET_ADMIN_USER,
+        "request_disaster_recovery",
+        (upgrader_api::RequestDisasterRecoveryInput {
+            module: station_wasm_module.clone(),
+            arg: Encode!(&station_api::SystemInstall::Init(station_api::SystemInit {
+                name: "Station".to_string(),
+                admins: vec![station_api::AdminInitInput {
```
mraszyk, line 506: Could we maybe have 3 admins here? Then we should see the quorum being set to 2 as we don't provide any quorum explicitly.

tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -418,6 +419,190 @@ fn test_disaster_recovery_flow() {
...
+                        assert_eq!(groups.len(), 1);
+                        assert_eq!(groups[0], group_id);
+                    }
+                    _ => {
+                        panic!("Unexpected approvers found");
+                    }
+                }
+            }
+            _ => {
+                panic!("Unexpected transfer policy found");
```
mraszyk, line 578:
```suggestion
                panic!("Unexpected request policy found");
```

tests/integration/src/disaster_recovery_tests.rs
mraszyk: Please ignore if this is already tested.

core/station/impl/src/services/system.rs
keplervital: just added a fail early inside the init hook itself, this way we don't need to check this here.

core/station/impl/src/errors/system.rs
```diff
@@ -10,6 +10,8 @@ pub enum SystemError {
     InitFailed { reason: String },
     #[error(r#"The canister needs at least one admin"#)]
     NoAdminsSpecified,
+    #[error(r#"There are too many admins defined, max is 255."#)]
```
mraszyk, line 13: > max is 255

the code checks for u16::MAX which is ~65k

core/station/impl/src/services/system.rs
```diff
@@ -626,4 +628,44 @@ mod tests {
...
+        assert_eq!(calc_initial_quorum(11, None), 6);
+        assert_eq!(calc_initial_quorum(12, None), 7);
+        assert_eq!(calc_initial_quorum(13, None), 7);
+        assert_eq!(calc_initial_quorum(14, None), 8);
+        assert_eq!(calc_initial_quorum(15, None), 8);
+        assert_eq!(calc_initial_quorum(16, None), 9);
+    }
+
+    #[test]
+    fn test_initial_quorum_is_custom() {
```
mraszyk, line 653: Could we have tests where Some(n) is
- 1
- half of the number of admins
- equal to the number of admins
- larger than the number of admins

mraszyk: This would be more useful than a lot of tests with Some(n) where n is always half of the number of admins.

core/station/impl/src/errors/system.rs
keplervital: i added the max as prop there, this way it's we defined closer to the check

mraszyk (approved):

PR #300, opened 2024-07-26
feat(wallet): prompt the user to update if new version available

keplervital:
This highlights to the user when a new version is available with an action button in the toolbar of the wallet.

The action button is only shown if there is no ongoing `Request` that is already asking to update the station or upgrader, also, for mobile i only added the icon instead of the `update` text to save on space.

Screenshot 2024-07-26 at 16 39 24


apps/wallet/src/components/change-canister/ChangeCanisterActionBtn.vue
```diff
@@ -135,19 +162,26 @@ const goToConfirmation = async (model: ChangeCanisterFormProps['modelValue']): P
 const submitUpgrade = async (model: ChangeCanisterFormProps['modelValue']): Promise<Request> => {
```
nathanosdev, line 162: This function is declared `async`,  but isn't using await.

```
const res = await station.service.changeCanister(
  // ...
);
station.checkVersionUpdates();

return res;
```

apps/wallet/src/stores/station.store.ts
```diff
@@ -419,15 +426,46 @@ export const useStationStore = defineStore('station', {
...
+            nextUpgraderVersion = dependency.version;
             break;
           }
         }
+
+        this.versionManagement.nextUpgraderVersion = nextUpgraderVersion;
+
+        // check if there is an existing request to update the canister, if so, we store the request id
+        // to avoid prompting the user to update again if they have already requested an update
+        this.service
```
nathanosdev, line 442: Style nit: this could also be converted to async/await syntax for consistency.

nathanosdev (approved):

apps/wallet/src/stores/station.store.ts
keplervital: Did that for the success scenario now, i am still using the catch there though, main reason is that if that call fails i don't want to throw to the overall try/catch block to prevent breaking the new version verification flow, also added a comment about that.

nathanosdev: A nested try/catch will have the same effect as using a `.catch` call nested inside a try/catch. Either way is fine for me anyway 👍

PR #303, opened 2024-07-30
feat(dfx-orbit): Initial dfx-orbit tool

Sawchord:
This PR adds an initial implementation of the `dfx-orbit` tool.
The `dfx-orbit` tool lives in `tools` subdirectory.

The PR also makes slight changes to the `integrations-test`, to support integration testing the tool.


tests/integration/src/dfx_orbit/assets.rs
```diff
@@ -0,0 +1,155 @@
...
+                        execution_method_cycles: None,
+                    },
+                ),
+                title: None,
+                summary: None,
+                execution_plan: None,
+            })
+            .await
+            .unwrap();
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
keplervital, line 123: any reason to sleep instead of doing a given number of ticks?

tests/integration/src/setup.rs
```diff
@@ -68,7 +68,7 @@ pub fn setup_new_env_with_config(config: SetupConfig) -> TestEnv {
         .with_nns_subnet()
         .with_application_subnet()
         .build();
-    env.set_time(SystemTime::now());
+    env.set_time(SystemTime::now() - Duration::from_secs(24 * 60 * 60));
```
keplervital, line 71: would be great to add a comment explaining the reason to put the time in the past

tests/integration/Cargo.toml
```diff
@@ -7,15 +7,21 @@ edition = '2021'
...
 hex = { workspace = true }
 orbit-essentials = { path = '../../libs/orbit-essentials', version = '0.0.2-alpha.3' }
 ic-cdk = { workspace = true }
 ic-ledger-types = { workspace = true }
 lazy_static = { workspace = true }
 num-bigint = { workspace = true }
 pocket-ic = { workspace = true }
+rand = { workspace = true }
+rand_chacha = { workspace = true }
+reqwest = { version = "0.12", default-features = false }
```
keplervital, line 20: nit: can we add this to the root Cargo.toml?

tests/integration/Cargo.toml
```diff
@@ -7,15 +7,21 @@ edition = '2021'
...
 sha2 = { workspace = true }
 serde = { workspace = true, features = ['derive'] }
+tempfile = "3.10"
```
keplervital, line 23: same here

tools/dfx-orbit/src/station_agent/error.rs
```diff
@@ -0,0 +1,21 @@
...
+use orbit_station_api::ApiErrorDTO;
+use thiserror::Error;
+
+pub type StationAgentResult<T> = Result<T, StationAgentError>;
+
+#[derive(Error, Debug)]
+pub enum StationAgentError {
+    #[error("The station API returned an error: {:?}", 0)]
+    Api(ApiErrorDTO),
+    #[error("Failed to parse canid: {0}")]
```
keplervital, line 11:
```suggestion
    #[error("Failed to parse candid: {0}")]
```

tools/dfx-orbit/src/station_agent.rs
```diff
@@ -0,0 +1,121 @@
...
+    // }
+
+    // pub fn update_canister(
+    //     &mut self,
+    //     canister: &str,
+    //     method_name: &str,
+    // ) -> anyhow::Result<UpdateBuilder> {
+    //     let canister_id = self.canister_id(canister)?;
+    //     Ok(self.agent().update(&canister_id, method_name))
+    // }
```
keplervital, line 83: nit: should this be removed?

tests/integration/src/dfx_orbit/assets.rs
Sawchord: I'll add a comment.

Here, the PocketIC is in "live" mode, so it uses realtime to advance the ticks.
If we leave life mode, I think we can no longer access the http enpoint to make the request.

I can add a comment. Also I wanted to figure out a way to switch back and forth between "live" mode and not live mode, but at the moment, we can only go live once.

tools/dfx-orbit/src/station_agent.rs
Sawchord: I can remove it for now. I am not entirely sure, whether I need this in the future, but I guess I can re-introduce this code, if I need it.

tools/dfx-orbit/src/dfx_extension_api.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// Gets the dfx_core interface
+    pub async fn dfx_interface(&mut self) -> anyhow::Result<DfxInterface> {
+        let network_name = self
+            .station_or_default(None)
+            .with_context(|| "Failed to get station")?
+            .network;
+        let interface_builder = DfxInterface::builder().with_network_named(&network_name);
+        let interface = interface_builder.build().await?;
+        if !interface.network_descriptor().is_ic {
```
keplervital, line 121: do you know how `is_ic` is determined? would this work if the `dfx.json` has a custom section of networks that also point at mainnet? e.g. staging, testing, production environments

tools/dfx-orbit/src/cli.rs
```diff
@@ -0,0 +1,93 @@
...
+            Ok(())
+        }
+        DfxOrbitSubcommands::Request(request_args) => {
+            let response = dfx_orbit
+                .station
+                .request(request_args.into_create_request_input(&dfx_orbit)?)
+                .await?;
+            let request_id = &response.request.id;
+            let request_url = dfx_orbit.station.request_url(request_id);
+            println!("Created request: {request_id}");
```
keplervital, line 37: nit: wasn't there a logger for this?

tools/dfx-orbit/src/dfx_extension_api.rs
Sawchord: If I understand correctly, it should work, as long as the proveder points to "icp0.io":
https://github.com/dfinity/sdk/blob/master/src/dfx-core/src/config/model/network_descriptor.rs#L80

keplervital: I see, ideally that dfx code would also check for `icp-api.io`, but for now i think it's fine then

keplervital (approved):

tools/dfx-orbit/src/cli.rs
Sawchord: So the whole output story is still a bit of a mess right now.
Right now, it uses a mix of logger and println.


Going forward, I would like to have the following design:
- **Logger**: Logs *internal* messages. A flag sets, how verbose the logging output will be. Goes to stderr
- **Output**: The output that is the result of the execution. This can be manipulated through flags (i.e. have json output). Goes to stdout

I have not had the time to properly implement that yet, but there is a whole overhaul planned. So for the time, I would leave it as is, if thats ok with you.

tests/integration/src/setup.rs
Sawchord: Added a comment.

tests/integration/Cargo.toml
Sawchord: Done

tools/dfx-orbit/src/dfx_extension_api.rs
Sawchord: Well, there are some upstream changes that I need to propose to the `sdk` team anyway. I can bring that up  too

tests/integration/Cargo.toml
Sawchord: Done

PR #304, opened 2024-07-30
feat(wallet): add disaster recovery config UI

olaszakos:
Adds new card on the Administration page if the user has SystemInfo privilege:
image

It'll show a Configure button, if the user has ChangeCanister privilege.

Opens the DR config dialog:
image

When DR is configured, it'll show the details:
image

It'll mark the DR user group on the User groups page, and will disable the delete button.
image

Todo:
- [x] add translation


apps/wallet/src/components/settings/DisasterRecoveryForm.vue
```diff
@@ -0,0 +1,89 @@
...
+  get: () => modelValue.value.quorum,
+  set: value => {
+    if (value && typeof value !== 'number') {
+      value = parseInt(value, 10);
+    }
+
+    modelValue.value.quorum = value;
+    modelValue.value = {
+      ...modelValue.value,
+    };
```
keplervital, line 78: nit: if there would be something deeply watching that value this change would avoid it to trigger the change twice.

```suggestion
    modelValue.value = {
      ...modelValue.value,
      quorum: value,
    };
```

keplervital (approved):
LGTM

apps/wallet/src/components/settings/DisasterRecoveryCard.vue
```diff
@@ -0,0 +1,155 @@
...
+          <VListItemSubtitle>
+            {{ systemInfo.disaster_recovery[0].user_group_name[0] || $t('terms.unknown') }}
+          </VListItemSubtitle>
+        </VListItem>
+        <VListItem class="px-0">
+          <VListItemTitle class="font-weight-bold"
+            >{{ $t(`terms.user_group_id`) }}:
+          </VListItemTitle>
+          <VListItemSubtitle>
+            {{ systemInfo.disaster_recovery[0].committee.user_group_id }}
```
keplervital, line 25: nit: would be nice to have here the copy to clipboard btn

olaszakos: added 👍

PR #305, opened 2024-07-30
feat(station): address book of external canisters

keplervital:
This adds the initial schema for managing external canisters, the controller methods and the new configure external canister operation are not yet implement.

**The new methods added are:**

- get_external_canister
- list_external_canisters
- get_external_canister_filters

Also the `CreateExternalCanisterOperationInput` was updated to support managing the newly created canisters and the `ConfigureExternalCanisterOperationInput` was added to enable users to manage those canisters after creation.


apps/wallet/src/generated/station/station.did
```diff
@@ -605,7 +605,63 @@ type ChangeExternalCanisterOperation = record {
...
-type CreateExternalCanisterOperationInput = record {};
+type CreateExternalCanisterOperationInput = record {
+  // By default, the canister id is created by the system.
+  //
+  // If provided, the canister must exist.
+  existing_canister_id : opt principal;
+  // The initial cycles to allocate to the canister.
+  //
+  // Ignored if the canister already exists.
+  initial_cycles : opt nat64;
```
olaszakos, line 616: What do you think about an `operation : Operation` variant 

AddExisting { canister_id : principal } 
CreateNew { initial_cycles : nat64 }

That way you can't represent conflicting input eg. providing both `existing_canister_id` and `initial_cycles`, it would need less documentation, and it would be extendable.

apps/wallet/src/generated/station/station.did
```diff
@@ -2083,6 +2150,169 @@ type CanisterStatusResponse = record {
...
+type GetExternalCanisterFiltersInput = record {
+  // Include the names of the external canisters in the result.
+  with_name : opt record {
+    // The prefix to use for filtering the names.
+    //
+    // If the prefix is not provided, any name will be returned.
+    prefix : opt text;
+  };
+  // Include the labels of the external canisters in the result.
+  with_labels : opt bool;
```
olaszakos, line 2301: any reason to use `opt bool` instead of just `bool`? Whats the difference between none and false?

apps/wallet/src/generated/station/station.did
olaszakos: was meaning to add comments to [/api/spec.did](https://github.com/dfinity/orbit/pull/305/files#diff-55b7ab3f382de0fb84e61ee21b84337109b8dc6e2e0a82ebb70f8440c9806c99)

keplervital: I like the approach you suggested, wdyt of something like the following:

```
type CreateExternalCanisterOperationKindCreateNew = record {
  // The initial cycles to allocate to the canister.
  //
  // If not set, only the minimal amount of cycles required to create the
  // canister will be allocated.
  initial_cycles : opt nat64;
};

type CreateExternalCanisterOperationKindAddExisting = record {
  // The canister id to use.
  canister_id : principal;
};

type CreateExternalCanisterOperationKind = variant {
  // A new canister is created.
  CreateNew : CreateExternalCanisterOperationKindCreateNew;
  // An existing canister is added to the station.
  AddExisting : CreateExternalCanisterOperationKindAddExisting;
};

type CreateExternalCanisterOperationInput = record {
  // The kind of create operation to perform.
  kind : CreateExternalCanisterOperationKind;
  // The name of the external canister.
  name : text;
  // The description of the external canister.
  description : opt text;
  // The labels of the external canister.
  labels : opt vec text;
  // What operations are allowed on the canister.
  permissions : ExternalCanisterPermissionsInput;
  // The request policies for the canister.
  request_policies : ExternalCanisterRequestPoliciesInput;
};
  ```

apps/wallet/src/generated/station/station.did
keplervital: adding and removing `opt` fields is always safe, so, this enables clients of this api that would not be providing the field to still work since the field is optional, same goes if we decide to remove the field in the future.

and this would make it consistent if we decide to add more fields to the record without breaking clients.

olaszakos: 👍

olaszakos (approved):

PR #306, opened 2024-08-05
feat(station): add external canister management

keplervital:
This initial PR adds the implementation of the external canister management.

**The following TODOs will be added in future PRs:**

- Integrate with request policies
- Integrate with permissions
- Use the service methods in the `ExternalCanisterController`
- Add integration tests for the top up and update ic settings operations
- Implement the search leveraging the indexes

### DIFF
--- a/core/station/impl/src/models/indexes/external_canister_index.rs
+++ b/core/station/impl/src/models/indexes/external_canister_index.rs
@@ -1,4 +1,5 @@
 use crate::models::ExternalCanisterId;
+use candid::Principal;
 use orbit_essentials::storable;
 
 /// The main index for external canisters.
@@ -14,6 +15,7 @@ pub struct ExternalCanisterIndex {
 #[storable]
 #[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum ExternalCanisterIndexKind {
+    CanisterId(Principal),
     Name(String),
     Label(String),
 }
--- a/core/station/impl/src/models/indexes/external_canister_index.rs
+++ b/core/station/impl/src/models/indexes/external_canister_index.rs
@@ -1,0 +1,25 @@
+use crate::models::ExternalCanisterId;
+use orbit_essentials::storable;
+
+/// The main index for external canisters.
+#[storable]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct ExternalCanisterIndex {
+    /// An indexed value of the external canister.
+    pub index: ExternalCanisterIndexKind,
+    /// The external canister id, which is a UUID.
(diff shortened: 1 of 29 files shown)


core/station/impl/src/models/indexes/external_canister_index.rs
```diff
@@ -0,0 +1,153 @@
...
+use orbit_essentials::storable;
+
+/// The main index for external canisters.
+#[storable]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct ExternalCanisterIndex {
+    /// An indexed value of the external canister.
+    pub index: ExternalCanisterIndexKind,
+    /// The external canister id, which is a UUID.
+    pub external_canister_id: ExternalCanisterId,
```
olaszakos, line 15: name this external_canister_resource_id for clarity?

core/station/impl/src/services/external_canister.rs
```diff
@@ -72,4 +128,357 @@ impl ExternalCanisterService {
...
+
+                self.check_unique_canister_id(&opts.canister_id, None)?;
+
+                opts.canister_id
+            }
+        };
+
+        let external_canister = ExternalCanisterMapper::from_create_input(canister_id, input);
+
+        external_canister.validate()?;
```
olaszakos, line 159: this can return an error when the canister is already created, losing it

core/station/impl/src/services/external_canister.rs
```diff
@@ -72,4 +128,357 @@ impl ExternalCanisterService {
...
+    pub async fn hard_delete_external_canister(
+        &self,
+        id: &ExternalCanisterId,
+    ) -> ServiceResult<ExternalCanister> {
+        let external_canister = self.get_external_canister(id)?;
+
+        // Deleting a canister requires the canister to be stopped first.
+        //
+        // See https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-delete_canister
+        if let Err((err_code, err_msg)) = stop_canister(CanisterIdRecord {
```
olaszakos, line 231: does this return an error if the canister is already stopped? maybe still go through with trying to delete the canister even if it errors out?

olaszakos (approved):

core/station/impl/src/services/external_canister.rs
keplervital: It could because we validate the name, description, labels, etc... but i moved the creation logic to after this validation, which means that to initially create the `ExternalCanister` struct i had to provide a temp principal, which should be fine since it is only persisted after the canister is created.

core/station/impl/src/services/external_canister.rs
keplervital: i think it would error out, i'll update to continue and just print the error if stop fails.

PR #307, opened 2024-08-06
feat(station): mint cycles to top up station

olaszakos:
Add way for canfund to top-up the canister it is running on - ObtainCycle trait + implementation.

First implementation: MintCycle which uses the Cycles Minting Canister to mint cycles from a user-defined Orbit ICP Account. For newly deployed stations this will be the first ICP account that is created from the frontend during deployment.


core/station/impl/src/services/system.rs
```diff
@@ -548,28 +582,41 @@ mod install_canister_handlers {
...
+                        .with_fallback_fund_cycles(250_000_000_000),
+                ));
+
+            if let Some(strategy) = cycle_obtain_strategy {
+                fund_manager_options = fund_manager_options
+                    .with_obtain_cycles(SYSTEM_SERVICE.get_obtain_cycle_config(&strategy));
+            }
+
+            fund_manager.with_options(fund_manager_options);
+            fund_manager.with_cycles_fetcher(Arc::new(FetchCyclesBalanceFromCanisterStatus {}));
```
keplervital, line 615: this would fail on `mainnet`, because the station is not a controller of itself and due to that it can't fetch it's own `canister_status`, would be great to add an integration test that validates this.

libs/canfund/src/api/cmc.rs
```diff
@@ -0,0 +1,162 @@
+use async_trait::async_trait;
```
keplervital, line 1: There's no dfinity crate with the logic of this file already right? if not, we could maybe put this on it's own lib and publish later as i imagine it would be helpful for other rust canisters.

libs/canfund/src/api/ledger.rs
```diff
@@ -0,0 +1,54 @@
...
+        async fn transfer(&self, _args: TransferArgs) -> CallResult<TransferResult> {
+            let mut locked = self.transfer_called_with.write().await;
+            locked.push(_args);
+
+            if let Some(value) = &self.returns_with {
+                return value.clone();
+            }
+
+            Ok(Ok(0))
+        }
```
keplervital, line 52:
```suggestion
        async fn transfer(&self, args: TransferArgs) -> CallResult<TransferResult> {
            let mut locked = self.transfer_called_with.write().await;
            locked.push(args);

            if let Some(value) = &self.returns_with {
                return value.clone();
            }

            Ok(Ok(0))
        }
```

libs/canfund/src/manager/mod.rs
```diff
@@ -193,7 +193,46 @@ impl FundManager {
 
             // Funds the canisters with the needed cycles.
             for (canister_id, needed_cycles) in canisters_to_fund {
-                if let Err((err_code, err_msg)) =
+                if canister_id == id() {
```
keplervital, line 196: atm the logic is only minting cycles if it's `self`, maybe this should be updated to the following:

```
if canister_id == id() or has_less_cycles_then_needed then mint
else send_cycles_from_host
```

libs/canfund/src/api/cmc.rs
olaszakos: Shouldn't that go to the nns monorepo?

core/station/impl/src/services/system.rs
olaszakos: See commit #[51cc892](https://github.com/dfinity/orbit/pull/307/commits/51cc8921647f5b465cfe8bf1a8784a45f918c25f)

libs/canfund/src/manager/mod.rs
olaszakos: My thinking here was that since the station is the first canister in the list, it'll get topped up first if needed, then the subsequent upgrader canister will get topped up successfully.

keplervital: This is true only for the Orbit case, once we release `canfund` as a crate this might not be the case anymore.

keplervital (approved):

PR #310, opened 2024-08-07
feat(station): integrate request policies and external canisters

keplervital:
This PR integrates the request policies with the external canisters.

**Note:** i've also removed the async need in the request policy service methods.

### DIFF
--- a/core/station/impl/src/services/external_canister.rs
+++ b/core/station/impl/src/services/external_canister.rs
@@ -1,21 +1,27 @@
 use super::permission::{PermissionService, PERMISSION_SERVICE};
+use super::request_policy::{RequestPolicyService, REQUEST_POLICY_SERVICE};
 use crate::core::ic_cdk::api::print;
 use crate::core::validation::EnsureExternalCanister;
 use crate::errors::ExternalCanisterError;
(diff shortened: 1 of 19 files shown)


core/station/impl/src/services/external_canister.rs
```diff
@@ -319,7 +469,29 @@ impl ExternalCanisterService {
...
+            .find_external_canister_call_permissions(&external_canister.canister_id)
+            .iter()
+            .for_each(|permission| {
+                self.permission_service
+                    .remove_permission(&permission.resource);
+            });
+
+        // Remove all request policies related to the external canister.
+        self.request_policy_repository
+            .find_external_canister_policies(&external_canister.canister_id)
```
olaszakos, line 483: does this remove the Read policies?

keplervital: there are no read request policies, those are only permissions, but you're right we still need to remove the read and change permissions here.

olaszakos: ah I confused myself 👍

olaszakos (approved):

PR #312, opened 2024-08-12
feat(station): add cache layer to permission repository

keplervital:
Introduces a cache layer to the`Permission` repository which can hold up to 300k permissions. 

Most noticeable benchmark improvement is `list_external_canisters_with_all_statuses`, which reduced the instruction cost from `3.5B` to `225M`.

### DIFF
--- a/core/station/impl/src/repositories/permission.rs
+++ b/core/station/impl/src/repositories/permission.rs
@@ -14,7 +14,7 @@ use candid::Principal;
 use ic_stable_structures::{memory_manager::VirtualMemory, StableBTreeMap};
 use lazy_static::lazy_static;
 use orbit_essentials::repository::Repository;
-use std::{cell::RefCell, collections::BTreeMap, sync::Arc};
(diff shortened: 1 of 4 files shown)


core/station/impl/src/repositories/permission.rs
```diff
@@ -34,26 +37,118 @@ pub struct PermissionRepository {}
...
+    /// up-to-date with the repository and that we have enough instructions to rebuild the cache.
+    pub fn build_cache(&self) {
+        if self.len() > Self::MAX_CACHE_SIZE {
+            print(format!(
+                "Only the first {} permissions will be added to the cache, the reposity has {} permissions.",
+                Self::MAX_CACHE_SIZE,
+                PERMISSION_REPOSITORY.len(),
+            ));
+
+            return;
```
olaszakos, line 135: This shouldn't return here, right?

olaszakos (approved):

PR #316, opened 2024-08-14
feat(station): add fee, comment, from address to csv export

olaszakos:


apps/wallet/src/mappers/requests.mapper.ts
```diff
@@ -597,9 +602,12 @@ export const mapRequestsToCsvTable = (
     details: 'Details',
   };
 
+  console.log('headers', headers);
```
keplervital, line 605:
```suggestion
```

keplervital (approved):

PR #318, opened 2024-08-16
feat(dfx-orbit): dfx-orbitv0.3

Sawchord:
This PR merges a new version of dfx-orbit.

Improvements:
- Proper logging, with settable verbosity (`v` and `q` flags). Logging goes to StdErr, output to StdOut
- Pass `--station [Name]` to address a station that is not the default station
- A `review` commands no longer by default dump the json to output, but rather provide a human readable Gist. The information is annotated with information, such as candid is parsed, and canister ids are reverse looked up to annotate the ids with the names of the canisters.


tools/dfx-orbit/src/args/review/list.rs
```diff
@@ -1,31 +1,43 @@
...
 
 impl From<ReviewListArgs> for ListRequestsInput {
     fn from(args: ReviewListArgs) -> Self {
-        let ReviewListArgs { only_approvable } = args;
         Self {
             requester_ids: None,
             approver_ids: None,
             statuses: None,
-            operation_types: None,
+            operation_types: (!args.all).then(external_canister_operations),
```
keplervital, line 31: Is there a reason why canister management is set as a default?

could we have those as the args of `ReviewListArgs`, maybe something like:

```
#[derive(Debug, Parser)]
pub struct ReviewListArgs {
    /// Filter the list by the request operation types provided.
    #[clap(short, long)]
    pub requests: Option<Vec<String>>,
    
    /// Filter the list by the request groups provided.
    /// 
    /// e.g. `canister_management` group includes `fund`, `call`, `create` and `configure` 
    /// external canister operations
    #[clap(short, long)]
    pub groups: Option<Vec<String>>,
    
   /// Filter only by `canister_management` operations.
   #[clap(short, long, conflicts_with = "groups,requests")]
   pub only_canister_management: Option<bool>, 
}
```

and for the requests and groups provide a command that would give the list of options to the user.

keplervital (approved):
LGTM

tools/dfx-orbit/src/args/review/list.rs
Sawchord: The reason that I think it makes sense to filter by `external_canisters` by default is that `dfx-orbit` is a tool to manage external canisters, at least at the moment. Other requests, such as managing  the users and groups is easier through the Web UI.

Therefore, I would expect that users would mostly care about external canisters at the moment. In the future, we might want to add other filters, similar to what you are proposing.

PR #319, opened 2024-08-16
feat(station): optimized repository lookups

keplervital:
It was identified that access to data in stable memory hurts performance due to the serialization/deserialization it needs, as a result these changes perform a major refactor on how data is stored and retrieved from stable memory.

**Updates:**

- Start to version the stable memory, past is now assumed to be `v0` and this one is `v1`
- Add a `migration.rs` that is now called as the first part of the `post_upgrade` hook
- Leverage `PocketIc` dump of stable memory to dump the memory of `v0` and perform e2e tests against it with the migration
- Renamed `ChangeCanister` to `SystemUpgrade` to avoid confusion with ChangeExternalCanister
- Manual implementation of serde::Deserialize to cover the changed data types
- Introduced the `StableDb`, `IndexedRepository` and `RebuildRespository` traits the migration less error prone and more standardize


libs/orbit-essentials/src/repository.rs
```diff
@@ -1,32 +1,110 @@
...
+pub trait Repository<Key, Value, Mem>: StableDb<Key, Value, Mem>
+where
+    Key: Eq + std::hash::Hash + Clone + Ord + Storable,
+    Value: Clone + Storable,
+    Mem: Memory,
+{
     /// Returns the list of records from the repository.
-    fn list(&self) -> Vec<Value>;
+    fn list(&self) -> Vec<Value> {
+        Self::with_db(|db| db.iter().map(|(_, v)| v.clone()).collect())
```
olaszakos, line 75: remove unnecessary clone()

core/station/impl/src/repositories/permission.rs
```diff
@@ -35,71 +34,80 @@ lazy_static! {
...
+                    permissions.push(Permission {
+                        resource: resource.clone(),
+                        allow: allow.clone(),
+                    });
+                });
+            });
+        } else {
+            Self::with_db(|db| {
+                db.iter().for_each(|(_, permission)| {
+                    permissions.push(permission.clone());
```
olaszakos, line 72: unnecessary clone

core/station/impl/src/repositories/user_group.rs
```diff
@@ -69,34 +82,26 @@ impl Repository<UUID, UserGroup> for UserGroupRepository {
...
-        let user_group_ids = self
-            .name_index
-            .find_by_criteria(UserGroupNameIndexCriteria {
-                name: name.to_string(),
-            });
+    /// Builds the cache from the stable memory repository.
+    ///
+    /// This method should only be called during init or upgrade hooks to ensure that the cache is
+    /// up-to-date with the repository and that we have enough instructions to rebuild the cache.
+    pub fn build_cache(&self) {}
```
olaszakos, line 99: is this empty on purpose?

keplervital: Nope, nice catch! Just added but the benchmarks showed no improvements based on it, but since this cache is small, we can keep it and have things consistent imo.

core/station/impl/src/repositories/request.rs
```diff
@@ -381,248 +232,149 @@ impl RequestRepository {
...
-                                as Box<dyn SelectionFilter<IdType = UUID>>
-                        })
-                        .collect(),
-                }),
-            }) as Box<dyn SelectionFilter<IdType = UUID>>;
-
-            filters.push(excludes_requester);
-        }
+            match ord {
+                std::cmp::Ordering::Equal => a_id.cmp(b_id),
```
olaszakos, line 368: handle reverse order also?

core/station/impl/src/repositories/request.rs
```diff
@@ -381,248 +232,149 @@ impl RequestRepository {
...
+                _ => match dir {
+                    station_api::SortDirection::Asc => ord.reverse(),
+                    station_api::SortDirection::Desc => ord,
+                },
+            }
+        });
 
-        filters
+        // self.sort_ids_with_strategy(&mut ids, &sort_by);
+        Ok(entries.iter().map(|(id, _)| *id).collect())
```
olaszakos, line 377: would *id return a copy? how about using into_iter here?

keplervital: Updated 👍

tests/integration/src/migration_tests.rs
```diff
@@ -0,0 +1,457 @@
...
+        &env,
+        canister_ids.station,
+        WALLET_ADMIN_USER,
+        USER_NR + 1, // +1 because there is the first admin user
+    );
+    assert_can_list_user_groups_endpoint(
+        &env,
+        canister_ids.station,
+        WALLET_ADMIN_USER,
+        USER_GROUPS_NR + 1, // +1 because there is the first admin user
```
olaszakos, line 85:
```suggestion
        USER_GROUPS_NR + 1, // +1 because there is the first admin group
```

tests/integration/src/migration_tests.rs
```diff
@@ -0,0 +1,457 @@
...
```
olaszakos, line 161:
```suggestion
        USER_GROUPS_NR + 1, // +1 because there is the first admin group
```

apps/wallet/src/composables/change-canister.composable.spec.ts
```diff
@@ -1,17 +1,14 @@
 import { describe, expect, it } from 'vitest';
```
olaszakos, line 1: this file was not renamed

libs/orbit-essentials/src/repository.rs
```diff
@@ -25,6 +25,11 @@ where
     /// Adds the indexes for the current value.
     fn add_entry_indexes(&self, value: &Value);
 
+    /// Clears all the indexes.
+    fn clear_indexes(&self) {
```
olaszakos, line 29: should this default implementation exist? one might forget to implement this for a repository and rust wont complain.

core/station/impl/src/repositories/user.rs
```diff
@@ -57,24 +56,90 @@ impl Default for UserRepository {
...
+            .for_each(|(index, id)| {
+                self.unique_index.insert(index, id);
+            });
+
+        entry.to_index_for_groups().into_iter().for_each(|index| {
+            self.group_status_index.insert(index);
+        });
+    }
+
+    fn clear_indexes(&self) {
```
olaszakos, line 98: clear cache?

core/station/impl/src/repositories/user_group.rs
```diff
@@ -25,20 +32,78 @@ lazy_static! {
...
+        entry
+            .to_unique_indexes()
+            .into_iter()
+            .for_each(|(index, id)| {
+                self.unique_index.insert(index, id);
+            });
+    }
+
+    /// Clears all the indexes for the user group.
+    fn clear_indexes(&self) {
```
olaszakos, line 67: clear cache?

PR #320, opened 2024-08-23
feat(station): rate limiter for request creation

mraszyk:
This PR adds a rate limiter for request creation into the station canister. The rate limiter logic is based on the corresponding NNS CMC canister [module](https://github.com/dfinity/ic/blob/master/rs/nns/cmc/src/limiter.rs). The rate limiter policy introduced by this PR is as follows:
- a single station user (or all principals without an associated station jointly) can create at most 2000 requests per 5 minutes;
- a single station user (or all principals without an associated station jointly) can create requests with at most 10MB of data per 5 minutes.

Worst-case heap memory consumption analysis:
- N: number of users (e.g., 100K)
- T: number of rate limiter types (fixed to 2 - count and size - in code)
- B: number of buckets per user and rate limiter type (fixed to 30 - aggregation window of 300s divided by bucket size of 10s - in code)
- S: data size per bucket (fixed to 16 - one u32 + one u64 + alignment - in code)
```
N x T x B x S = 100,001 * 2 * 30 * 16 ~ 92M
```
So even with 100K users who would all create one request every 10s for at least five minutes, we'd only get to around 92M of heap memory used by the rate limiter.


core/station/impl/src/mappers/rate_limiter.rs
```diff
@@ -0,0 +1,20 @@
+use crate::models::rate_limiter::RequestRateLimiterType;
+use station_api::RequestOperationInput;
+
+impl From<&RequestOperationInput> for RequestRateLimiterType {
+    fn from(input: &RequestOperationInput) -> RequestRateLimiterType {
+        match input {
+            RequestOperationInput::Transfer(_) => RequestRateLimiterType::Transfer,
+            _ => RequestRateLimiterType::Misc,
```
keplervital, line 8: Should the Misc key be `RequestRateLimiterType::Misc` or `RequestRateLimiterType::Misc(type)`? 

where type would be the specific request operation type, so that we would have `N` requests per type and not `N` requests over all types

mraszyk: Having a finer granularity would increase the memory footprint by 11.5x (as we'd go from 2 variants to 23 variants). But I guess it depends on the requirements if we need finer granularity.

keplervital (approved):
Great Job 💯

PR #322, opened 2024-08-24
fix(station): requests are visible to users with approval rights

keplervital:
It was noticed that once a request is not in the `Created` status anymore, if the user was not the requester or one of the approvers they would not have access to the request, however, if a user has approval rights, they should still be able to see the request.

This PR fixes the above described issue, also, it fixes the UI to show the recently renamed ChangeCanister to the SystemUpgrade permission.


.gitignore
```diff
@@ -71,3 +71,7 @@ wasms/*.wasm.gz
 .canbench_output
 .nx/cache
 /artifacts
+
+# The latest memory dump of the station will change every time the tests are run due to 
+# the UUID being generated with a random seed. We don't need to track this file.
```
mraszyk, line 76: > the UUID being generated with a random seed

the seed is derived deterministically in tests (it doesn't depend on system time etc.)

mraszyk: What makes the memory non-deterministic are the time-stamps stored in various records which are derived from the current time of PocketIC which is set to now: https://github.com/dfinity/orbit/blob/ecd210dbca70e69235462a102c30d4a40bf74309/tests/integration/src/setup.rs#L95

olaszakos (approved):

PR #323, opened 2024-08-25
chore(station): add authorization middleware to me and notifications endpoints

mraszyk:
This PR adds the authorization middleware to the "me" and notification endpoints. The only endpoints without the authorization middleware after this PR are "health_status" (very simple and cheap) and "http_request" (supposed to be publicly accessible).

PR #327, opened 2024-08-27
chore(station): remove fn assert_notification_access

mraszyk:
This PR removes notification access checks in the notification service since the access checks are now performed early in the authorization middleware (cf. https://github.com/dfinity/orbit/pull/323).


keplervital (approved):
LGTM

PR #328, opened 2024-08-27
chore(station): introduce candid decoding quota for http_request

mraszyk:
This PR adds candid decoding quota to `http_request` endpoint of the station canister. This way, excessively large HTTP requests are rejected early.


tests/integration/src/http.rs
```diff
@@ -0,0 +1,53 @@
+use crate::setup::setup_new_env;
+use crate::TestEnv;
+use candid::{Decode, Encode, Principal};
+use orbit_essentials::api::{HttpRequest, HttpResponse};
+use pocket_ic::{PocketIc, WasmResult};
+
+fn test_candid_decoding_quota(env: &PocketIc, canister_id: Principal) {
```
keplervital, line 7: any reason not to inline this inside `test_http_request_deconding_quota`?

mraszyk: In a follow up PR, I'll introduce the decoding quota into the control panel canister. I didn't do it in this PR so that its scope can be chore(station).

keplervital (approved):

PR #331, opened 2024-08-27
feat(station): notify failed station upgrade

mraszyk:
This PR makes the upgrader canister notify the station canister about a failed station upgrade. Consequently, the station can mark the (so far processing) station upgrade request as failed.

Note that the `trigger_upgrade` method of the upgrader needs to perform the upgrade _after_ responding to the station so that the station can be stopped. But this means that the response of the upgrader to the station cannot contain any subsequent failures encountered during the actual upgrade. Hence, such failures need to be communicated in a separate call from the upgrader to the station.

### DIFF
--- a/core/station/impl/src/services/system.rs
+++ b/core/station/impl/src/services/system.rs
@@ -24,7 +24,7 @@ use crate::{
     },
     SYSTEM_VERSION,
 };
-use candid::{CandidType, Principal};
+use candid::Principal;
 use canfund::{
     api::{cmc::IcCyclesMintingCanister, ledger::IcLedgerCanister},
     manager::options::ObtainCyclesOptions,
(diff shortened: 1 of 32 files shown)


core/station/impl/src/services/system.rs
```diff
@@ -136,13 +132,19 @@ impl SystemService {
...
+        arg: &[u8],
+    ) -> ServiceResult<()> {
         let upgrader_canister_id = self.get_upgrader_canister_id();
 
         ic_cdk::call(
             upgrader_canister_id,
             "trigger_upgrade",
-            (InstallCanisterParams {
+            (UpgradeParams {
+                request_id: Uuid::from_bytes(request_id).hyphenated().to_string(),
```
keplervital, line 147: the upgrade request id is already stored in the station, under `https://github.com/dfinity/orbit/blob/main/core/station/impl/src/models/system.rs#L47` instead of adding this in the UpgradeParams the above `notify_failed_station_upgrade` method would just need to receive the `reason`

mraszyk: Good point! Fixed.

keplervital (approved):
LGTM

PR #333, opened 2024-08-28
chore: certify that http_request endpoint skips certification in control panel and station

mraszyk:
This PR certifies that asset certification is skipped in the control panel and station canisters so that the canisters' `http_request` endpoint can be accessed without using the raw endpoint.

### DIFF
--- a/core/station/impl/src/controllers/http.rs
+++ b/core/station/impl/src/controllers/http.rs
@@ -2,12 +2,15 @@ use crate::{core::ic_cdk::api::canister_balance, SERVICE_NAME};
 use ic_cdk_macros::query;
 use lazy_static::lazy_static;
 use orbit_essentials::api::{HeaderField, HttpRequest, HttpResponse};
+use orbit_essentials::http::add_skip_certification_headers;
 use orbit_essentials::metrics::with_metrics_registry;
 
 // Canister entrypoints for the controller.
 #[query(name = "http_request", decoding_quota = 10000)]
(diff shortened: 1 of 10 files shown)


keplervital (approved):

core/station/impl/src/controllers/http.rs
```diff
@@ -2,12 +2,15 @@ use crate::{core::ic_cdk::api::canister_balance, SERVICE_NAME};
...
 use orbit_essentials::api::{HeaderField, HttpRequest, HttpResponse};
+use orbit_essentials::http::add_skip_certification_headers;
 use orbit_essentials::metrics::with_metrics_registry;
 
 // Canister entrypoints for the controller.
 #[query(name = "http_request", decoding_quota = 10000)]
 async fn http_request(request: HttpRequest) -> HttpResponse {
-    CONTROLLER.router(request).await
+    let mut resp = CONTROLLER.router(request).await;
+    add_skip_certification_headers(&mut resp);
```
keplervital, line 12: this is failing is tests because it's calling the actual `set_certified_data` which is only available in a canister environment, and using `orbit_essentials::http::add_skip_certification_headers` the #[cfg(test)] does not apply across dependencies. for this to work you'll likely need to pass the `set_certified_data` as a callback and get that by importing in the control-panel and station with `crate::core::ic_cdk::api::set_certified_data`

mraszyk: Let me then just return the certified data to be set from the library and invoke the system API here.

PR #340, opened 2024-09-05
feat(station): asset crud operations

olaszakos:


core/station/impl/src/repositories/asset.rs
```diff
@@ -175,6 +175,12 @@ impl AssetRepository {
             });
         });
     }
+
+    pub fn exists_unique(&self, blockchain: &str, symbol: &str) -> Option<AssetId> {
+        let key = UniqueIndexKey::AssetSymbolBlockchain(symbol.to_string(), blockchain.to_string());
```
keplervital, line 180: should the symbol be lowercased?

core/station/impl/src/services/asset.rs
```diff
@@ -0,0 +1,303 @@
...
+            .get(asset_id)
+            .ok_or(AssetError::NotFound {
+                id: Uuid::from_bytes(*asset_id).hyphenated().to_string(),
+            })?;
+
+        Ok(asset)
+    }
+
+    pub async fn create(&self, input: AddAssetOperationInput) -> ServiceResult<Asset> {
+        let id = generate_uuid_v4().await;
```
keplervital, line 56: now that we have the `Uuid::new_v4` being seeded you could potentially use that here and remove the async need of this method

core/station/impl/src/models/asset.rs
```diff
@@ -102,11 +109,31 @@ fn validate_decimals(decimals: u32) -> ModelValidatorResult<AssetError> {
...
+
+    Ok(())
+}
+
 impl ModelValidator<AssetError> for Asset {
     fn validate(&self) -> ModelValidatorResult<AssetError> {
         validate_symbol(&self.symbol)?;
         validate_name(&self.name)?;
         validate_decimals(self.decimals)?;
+        validate_uniqueness(&self.id, &self.symbol, &self.blockchain)?;
```
keplervital, line 136: would be nice to extend the tests to make sure this is covered

keplervital (approved):
LGTM

PR #341, opened 2024-09-05
feat(station): serve metrics as properly certified HTTP asset

mraszyk:
This PR serves station metrics as a properly certified HTTP asset. The metrics include the timestamp at which they were exported. The metrics are exported every time the metrics change, but they are not exported if only the canister cycles balance changes.

This PR is currently blocked on installing canisters via the chunk API as the joint size of the station and upgrade canister WASMs exceed 2MiB.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2135,6 +2135,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ic-asset-certification"
+version = "2.6.0"
+source = "git+https://github.com/dfinity/response-verification?rev=eefc99e2a0041ae46873864e6e2f7aa8506a361f#eefc99e2a0041ae46873864e6e2f7aa8506a361f"
+dependencies = [
+ "globset",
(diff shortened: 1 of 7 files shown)


core/control-panel/impl/src/controllers/http.rs
```diff
@@ -1,115 +1,97 @@
...
+            .gauge_mut(
+                "canister_cycles_balance",
+                "cycles balance available to the canister",
+            )
+            .set(canister_balance() as f64);
+    });
+    with_metrics_registry(SERVICE_NAME, |registry| {
+        registry
+            .gauge_mut(
+                "metrics_timestamp",
```
keplervital, line 55: Alternatively the metrics registry could be adapted to use the prometheus format which optionally can include timestamp per each metric (check [here](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-format-details))

mraszyk: I don't see how this can be done with the Rust prometheus crate.

PR #343, opened 2024-09-06
feat(wallet): add external canister setup wizard

keplervital:
This PR adds the creation wizard of external canisters.

Also, i've created 2 components that should be reused in other contexts:

- `SuggestiveAutocomplete`: Can be used to search, select and create on tags, we use it for labels in the external canister context.
- `CyclesInput`: Validates and provides some of the available units for cycles `TC`, `Billion`, `Million`, `e8s`

**Here's some screenshots of the wizard:**

Screenshot 2024-09-06 at 16 58 55
Screenshot 2024-09-06 at 16 59 22
Screenshot 2024-09-06 at 16 59 35


apps/wallet/src/utils/helper.utils.ts
```diff
@@ -290,3 +290,17 @@ export const parseLocationQuery = (query: LocationQuery): Record<string, string[
...
+
+export const parseToBigIntOrUndefined = (
+  value: string | undefined | bigint,
+): bigint | undefined => {
+  if (value === undefined || value === null) {
+    return undefined;
+  }
+
+  if (typeof value === 'string') {
+    return value.trim() !== '' ? BigInt(value) : undefined;
```
olaszakos, line 302: this can throw an exception https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/BigInt

keplervital: True, fixed and added some tests.

olaszakos (approved):

PR #346, opened 2024-09-12
feat(wallet): add canister top up dialog

keplervital:
Adds the top-up canister dialog as shown in the screenshot:

Screenshot 2024-09-12 at 12 08 52

**fyi:** The request for now will fail because we are not yet exposing the configuration of the `FundExternalCanister` policy, this will come in follow up PRs.


apps/wallet/src/components/inputs/CyclesInput.vue
```diff
@@ -7,8 +7,15 @@
       :label="props.label"
       :variant="props.variant"
       :density="props.density"
+      :readonly="props.readonly"
       type="number"
-      :rules="[intNumberRangeRule(props.label ?? 'cycles', 1, Number.MAX_SAFE_INTEGER)]"
+      :hint="props.hint"
+      :persistent-hint="!!props.hint"
+      :rules="
```
olaszakos, line 14: How about something like 
```[...(props.required ? [requiredRule] : []), intNumberRangeRule(props.label ?? 'cycles', 1, Number.MAX_SAFE_INTEGER)]```
to not repeat the intNumberRangeRule rule?

olaszakos (approved):

PR #348, opened 2024-09-16
feat(station): add ability to cancel pending requests to EditUserOperation

mraszyk:
This PR adds the ability to cancel pending requests (with request status `Created`) to `EditUserOperation` so that, e.g., admins can cancel pending requests of users when they make them inactive.


core/station/impl/src/repositories/request.rs
```diff
@@ -387,6 +388,19 @@ impl RequestRepository {
         Ok(entries.into_iter().map(|(id, _)| id).collect())
     }
 
+    pub async fn cancel_request(
```
keplervital, line 391: nit: potentially async could be removed here

keplervital (approved):

core/station/impl/src/repositories/request.rs
mraszyk: Good point! Done.

PR #349, opened 2024-09-17
chore(ci): more test threads for integration tests

mraszyk:


mraszyk:
Tweaking the number of test threads doesn't really help to speed up integration tests.

PR #350, opened 2024-09-17
feat(wallet): add canister unlink dialog

keplervital:
Adds the unlink canister functionality to the user interface.

**Screenshoot:**

Screenshot 2024-09-17 at 11 54 37

### DIFF
--- a/apps/wallet/src/locales/en.locale.ts
+++ b/apps/wallet/src/locales/en.locale.ts
@@ -425,6 +425,8 @@ export default {
     top_up: 'Top Up',
     configuration: 'Configuration',
     unlink: 'Unlink',
+    unlink_title: 'Unlink Canister',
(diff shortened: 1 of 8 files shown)


apps/wallet/src/locales/en.locale.ts
```diff
@@ -425,6 +425,8 @@ export default {
     top_up: 'Top Up',
     configuration: 'Configuration',
     unlink: 'Unlink',
+    unlink_title: 'Unlink Canister',
+    unlink_soft_delete: "Keep the canister on the Intenret Computer, only removes it's reference.",
```
olaszakos, line 429:
```suggestion
    unlink_soft_delete: "Keep the canister on the Internet Computer, only remove its reference.",
```

olaszakos (approved):

PR #358, opened 2024-09-26
chore(station): execute requests in separate update calls

mraszyk:
This PR makes the station canister execute requests in separate update calls so that traps in request execution can be "caught" individually (per request) as reject responses from those separate update calls. Previously, a trap in a request execution might trap the job driving request execution resulting in that job being stuck indefinitely and consequently no requests being able to execute.


keplervital (approved):
LGTM

PR #359, opened 2024-09-27
feat(wallet): add canister install dialog

keplervital:
Adds the install canister wasm module dialog and form.

Screenshot 2024-09-27 at 10 07 26


apps/wallet/src/components/inputs/CanisterArgumentField.vue
```diff
@@ -0,0 +1,145 @@
...
+    if (typeof value !== 'string') {
+      throw new Error(i18n.t('external_canisters.wasm_args_invalid_format'));
+    }
+
+    let rawArgument = value.trim();
+    let parsedArgument: Uint8Array;
+
+    switch (selectedParseFormat.value) {
+      case 'hex':
+        parsedArgument = new Uint8Array(hexStringToArrayBuffer(rawArgument));
```
olaszakos, line 122: if I put invalid characters in the hex string, they will be replaced with 0's in the parsed result without throwing an error

olaszakos (approved):

apps/wallet/src/components/inputs/CanisterArgumentField.vue
keplervital: i see, i'll add some validation for that

PR #360, opened 2024-09-28
fix(station): bound the maximum number of processing requests

mraszyk:
This PR bounds the maximum number of processing requests to 400. Having more than 500 processing requests could lead to spurious failures since the queue capacity between any pair of canisters is 500 and thus, e.g., there cannot be more than 500 ICP ledger transfers processing concurrently. With a value of 400, we also leave some slack.


keplervital (approved):
LGTM

PR #361, opened 2024-09-30
fix(http): improve route matching to fix mac subdomain resolution issue

jedna:
Safari and shell interpreters on MacOS don't implicitly resolve subdomains on localhost (e.g. container_id.localhost). This behavior causes an issue with one of our integration tests. Proposed change updates the test to use localhost?canisterId=X schema instead and updates the router matching logic to allow for such change.


libs/orbit-essentials/src/http.rs
```diff
@@ -61,3 +61,37 @@ fn cbor_encode(value: &impl Serialize) -> Vec<u8> {
...
+
+pub fn not_found() -> HttpResponse {
+    HttpResponse {
+        status_code: 404,
+        headers: vec![HeaderField("Content-Type".into(), "text/plain".into())],
+        body: "404 Not Found".as_bytes().to_owned(),
+    }
+}
+
+pub fn parse_path(url: &str) -> Option<&str> {
```
mraszyk, line 73: Why are you not using a library for this?

jedna: We discussed this with @keplervital and concluded it's unnecessary to include a library for such a simple and isolated use case where only a dev environment is affected. WDYT?

jedna: > are you sure that none of the existing dependencies can deal with that?
Good call, I'm not. I'll investigate.

core/control-panel/impl/src/controllers/http.rs
```diff
@@ -35,19 +35,14 @@ impl HttpController {
     }
 
     async fn router(&self, request: HttpRequest) -> HttpResponse {
-        if request.url == "/metrics" || request.url == "/metrics/" {
-            return self.metrics(request).await;
+        match parse_path(&request.url) {
+            Some(url) => match url {
+                "/metrics" | "/metrics/" => self.metrics(request).await,
+                "/metrics/sd" | "/metrics/sd/" => self.metrics_service_discovery(request).await,
```
keplervital, line 41:
```suggestion
            Some(path) => match path.trim_end_matches('/') {
                "/metrics" => self.metrics(request).await,
                "/metrics/sd" => self.metrics_service_discovery(request).await,
```

core/station/impl/src/controllers/http.rs
```diff
@@ -27,15 +27,13 @@ impl HttpController {
     }
 
     async fn router(&self, request: HttpRequest) -> HttpResponse {
-        if request.url == "/metrics" || request.url == "/metrics/" {
-            return self.metrics(request).await;
+        match parse_path(&request.url) {
+            Some(url) => match url {
+                "/metrics" | "/metrics/" => self.metrics(request).await,
```
keplervital, line 32:
```suggestion
            Some(path) => match path.trim_end_matches('/') {
                "/metrics" => self.metrics(request).await,
```

libs/orbit-essentials/src/http.rs
```diff
@@ -61,3 +61,37 @@ fn cbor_encode(value: &impl Serialize) -> Vec<u8> {
...
+    fn test_parse_path() {
+        assert_eq!(parse_path("/long/path?param"), Some("/long/path"));
+        assert_eq!(parse_path("/path?query=1"), Some("/path"));
+        assert_eq!(parse_path("/path#unwanted"), Some("/path#unwanted"));
+        assert_eq!(parse_path("/"), Some("/"));
+        assert_eq!(parse_path(""), Some(""));
+    }
+
+    #[test]
+    fn test_cbor_encode() {
```
keplervital, line 91: what is this test covering?

jedna: It's a dummy test for the `cbor_encode` function from this module, which had no test. I was mostly curious about the output, so I left it there. I can remove it if it's useless.

libs/orbit-essentials/src/http.rs
mraszyk: I see, thank you for looking into this!

libs/orbit-essentials/src/http.rs
keplervital: i see, let's keep then, but maybe remove the `println!`

keplervital (approved):

PR #366, opened 2024-09-30
feat(station): external canister input to accept opt policies and permissions by type

keplervital:
With the existing API it was not possible to optionally only update the ExternalCanister change or call request policies, we always had to provide both, same with the permissions for read, change, call. This PR updates that and marks the individual policy and permission types as optional.

### DIFF
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -2318,10 +2318,10 @@ type ExternalCanisterChangeCallPermissionsInput = variant {
   // Replaces all the call permissions with the provided list, if the list is empty
   // all the call permissions will be removed.
   ReplaceAllBy : vec ExternalCanisterCallPermission;
-  // Override the call permissions from the specified methods.
-  OverrideSpecifiedByMethods : vec ExternalCanisterCallPermission;
-  // Remove call permissions of the specified methods.
-  RemoveByMethods : vec text;
+  // Override the call permissions from the specified execution methods.
+  OverrideSpecifiedByExecutionMethods : vec ExternalCanisterCallPermission;
+  // Remove call permissions of the specified execution methods.
+  RemoveByExecutionMethods : vec text;
 };
 
 // The input type for setting the permissions for the external canister.
@@ -2358,8 +2358,8 @@ type ExternalCanisterChangeCallRequestPoliciesInput = variant {
   ReplaceAllBy : vec ExternalCanisterCallRequestPolicyRuleInput;
   // Remove call request policies by the provided ids.
   RemoveByPolicyIds : vec UUID;
-  // Override the request policies for the specified methods.
-  OverrideSpecifiedByMethods : vec ExternalCanisterCallRequestPolicyRuleInput;
+  // Override the request policies for the specified execution methods.
+  OverrideSpecifiedByExecutionMethods : vec ExternalCanisterCallRequestPolicyRuleInput;
 };
 
 // The input type for setting the request policies for an existing external canister.
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -653,7 +653,7 @@ type CreateExternalCanisterOperationInput = record {
   // The labels of the external canister.
   labels : opt vec text;
   // What operations are allowed on the canister.
-  permissions : ExternalCanisterPermissionsInput;
+  permissions : ExternalCanisterPermissionsCreateInput;
   // The request policies for the canister.
   request_policies : ExternalCanisterRequestPoliciesInput;
 };
@@ -665,7 +665,7 @@ type ConfigureExternalCanisterSettingsInput = record {
   // The labels of the external canister.
   labels : opt vec text;
   // What operations are allowed on the canister.
-  permissions : opt ExternalCanisterPermissionsInput;
+  permissions : opt ExternalCanisterPermissionsUpdateInput;
   // The request policies for the canister.
   request_policies : opt ExternalCanisterRequestPoliciesInput;
   // The state of the external canister.
@@ -2310,8 +2310,21 @@ type ExternalCanisterPermissions = record {
   calls : vec ExternalCanisterCallPermission;
 };
 
+// The create input type for setting the permissions for the external canister.
+type ExternalCanisterPermissionsCreateInput = ExternalCanisterPermissions;
+
 // The input type for setting the permissions for the external canister.
-type ExternalCanisterPermissionsInput = ExternalCanisterPermissions;
+type ExternalCanisterPermissionsUpdateInput = record {
+  // Who can read information about the canister (e.g. canister status),
+  // changes to this permission can be made by the `change` permission.
+  read : opt Allow;
+  // Who can make changes to the canister, includes:
+  // - changing the permissions
+  // - install operations
+  change : opt Allow;
+  // The permissions for the calling methods on the canister.
+  calls : opt vec ExternalCanisterCallPermission;
+};
 
 // The request policy rules for the external canister.
 type ExternalCanisterRequestPolicies = record {
@@ -2324,9 +2337,9 @@ type ExternalCanisterRequestPolicies = record {
 // The input type for setting the request policies for the external canister.
 type ExternalCanisterRequestPoliciesInput = record {
   // The request policy rules for the canister change operation.
-  change : vec ExternalCanisterChangeRequestPolicyRuleInput;
+  change : opt vec ExternalCanisterChangeRequestPolicyRuleInput;
   // The request policy rules for the calling methods on the canister.
-  calls : vec ExternalCanisterCallRequestPolicyRuleInput;
+  calls : opt vec ExternalCanisterCallRequestPolicyRuleInput;
 };
 
 // An external canister that the station can interact with.
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -2322,7 +2322,7 @@ type ExternalCanisterPermissionsUpdateInput = record {
   // - changing the permissions
   // - install operations
   change : opt Allow;
-  // The permissions for the calling methods on the canister.
+  // The permissions for calling methods on the canister.
   calls : opt vec ExternalCanisterCallPermission;
 };
 
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -655,7 +655,7 @@ type CreateExternalCanisterOperationInput = record {
   // What operations are allowed on the canister.
   permissions : ExternalCanisterPermissionsCreateInput;
   // The request policies for the canister.
-  request_policies : ExternalCanisterRequestPoliciesInput;
+  request_policies : ExternalCanisterRequestPoliciesCreateInput;
 };
 
 type ConfigureExternalCanisterSettingsInput = record {
@@ -667,7 +667,7 @@ type ConfigureExternalCanisterSettingsInput = record {
   // What operations are allowed on the canister.
   permissions : opt ExternalCanisterPermissionsUpdateInput;
   // The request policies for the canister.
-  request_policies : opt ExternalCanisterRequestPoliciesInput;
+  request_policies : opt ExternalCanisterRequestPoliciesUpdateInput;
   // The state of the external canister.
   state : opt ExternalCanisterState;
 };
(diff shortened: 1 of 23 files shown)


core/station/api/spec.did
```diff
@@ -2310,8 +2310,21 @@ type ExternalCanisterPermissions = record {
...
-type ExternalCanisterPermissionsInput = ExternalCanisterPermissions;
+type ExternalCanisterPermissionsUpdateInput = record {
+  // Who can read information about the canister (e.g. canister status),
+  // changes to this permission can be made by the `change` permission.
+  read : opt Allow;
+  // Who can make changes to the canister, includes:
+  // - changing the permissions
+  // - install operations
+  change : opt Allow;
+  // The permissions for the calling methods on the canister.
```
mraszyk, line 2325:
```suggestion
  // The permissions for calling methods on the canister.
```

core/station/api/src/external_canister.rs
```diff
@@ -43,10 +40,24 @@ pub struct CreateExternalCanisterOperationInput {
     pub name: String,
     pub description: Option<String>,
     pub labels: Option<Vec<String>>,
-    pub permissions: ExternalCanisterPermissionsInput,
+    pub permissions: ExternalCanisterPermissionsCreateInput,
     pub request_policies: ExternalCanisterRequestPoliciesInput,
 }
 
+#[derive(CandidType, serde::Serialize, Deserialize, Debug, Clone)]
+pub struct ExternalCanisterPermissionsCreateInput {
```
mraszyk, line 48: Why does this have to be defined from scratch?

core/station/impl/src/services/external_canister.rs
```diff
@@ -598,119 +608,133 @@ impl ExternalCanisterService {
...
-                            external_canister.canister_id.to_text()
-                        ));
+        // manage the policies for the `Call` requests
+        if let Some(updated_call_policies) = input.calls {
+            let policies_to_remove = current_policies
+                .calls
+                .iter()
+                .filter(|policy_id| {
+                    !updated_call_policies
+                        .iter()
```
mraszyk, line 674: This nested iter takes quadratic time.

core/station/impl/src/services/external_canister.rs
```diff
@@ -598,119 +608,133 @@ impl ExternalCanisterService {
...
-                    if !current_policies.contains(&policy_id) {
-                        print(format!(
-                            "Policy with id {} not found for external canister {}",
-                            Uuid::from_bytes(policy_id).hyphenated(),
-                            external_canister.canister_id.to_text()
-                        ));
+            for updated_change_policy in updated_change_policies {
+                match updated_change_policy.policy_id {
+                    Some(policy_id) => {
+                        if !current_policies.change.contains(&policy_id) {
```
mraszyk, line 635: Why are we continuing here instead of erroring out?

core/station/impl/src/services/external_canister.rs
```diff
@@ -598,119 +608,133 @@ impl ExternalCanisterService {
...
-                Some(policy_id) => {
-                    if !current_policies.contains(&policy_id) {
-                        print(format!(
-                            "Policy with id {} not found for external canister {}",
-                            Uuid::from_bytes(policy_id).hyphenated(),
-                            external_canister.canister_id.to_text()
-                        ));
+        // manage the policies for the `Call` requests
+        if let Some(updated_call_policies) = input.calls {
+            let policies_to_remove = current_policies
```
mraszyk, line 669: current_policies includes both change and calls so here we'd be removing all change policies (and vice-versa above), right?

mraszyk:
Even after this PR, if we provide some call policies, then we need to provide all of them (as policies missing in the input are gonna be deleted). Is this intended?

core/station/api/src/external_canister.rs
keplervital: can you be more specific?

mraszyk: In the candid spec, there's a type synonym:
```
// The create input type for setting the permissions for the external canister.
type ExternalCanisterPermissionsCreateInput = ExternalCanisterPermissions;
```
but here we define the type fully.

core/station/impl/src/services/external_canister.rs
```diff
@@ -686,13 +683,9 @@ impl ExternalCanisterService {
...
                     Some(policy_id) => {
                         if !current_policies.calls.contains(&policy_id) {
-                            print(format!(
-                                "Policy with id {} not found for external canister {}",
-                                Uuid::from_bytes(policy_id).hyphenated(),
-                                external_canister.canister_id.to_text()
-                            ));
-
-                            continue;
+                            return Err(ExternalCanisterError::NotFound {
```
mraszyk, line 686: Do we need to return here as editing a non-existing policy later would fail anyway.

keplervital: ah right, missed that, removed then 👍

core/station/api/src/external_canister.rs
keplervital: i see, makes sense, i did the name on the impl which now uses `ExternalCanisterPermissionsDTO`

keplervital:
>Even after this PR, if we provide some call policies, then we need to provide all of them (as policies missing in the input are gonna be deleted). Is this intended?

As discussed, will update the input for updating calls of policies and permissions to have a similar behaviour as the ChangeMetadata type

core/station/impl/src/repositories/indexes/request_policy_resource_index.rs
```diff
@@ -158,6 +191,49 @@ impl RequestPolicyResourceIndexRepository {
...
+                            CallExternalCanisterResourceTarget {
+                                execution_method: ExecutionMethodResourceTarget::ExecutionMethod(
+                                    CanisterMethod { canister_id: id, method_name: method }
+                                ),
+                                ..
+                            }
+                        ))
+                        if id == canister_id && execution_method == method
+                    )
+                })
```
mraszyk, line 232: Isn't this fragile? It seems to depend on the fact that `ValidationMethodResourceTarget::No` is the least value of the `ValidationMethodResourceTarget` type.

keplervital: It does depend on that, we use this in other places as well with range queries. To make this more resilient we could instead of deriving `[(PartialOrd, Ord)]` we could implement that manually, but i don't think we need to do it for now.

mraszyk: > i don't think we need to do it for now

If we have a test, we should fine actually.

core/station/impl/src/repositories/indexes/request_policy_resource_index.rs
```diff
@@ -248,4 +324,34 @@ mod tests {
...
+            let index = RequestPolicyResourceIndex {
+                resource: Resource::ExternalCanister(ExternalCanisterResourceAction::Call(
+                    CallExternalCanisterResourceTarget {
+                        execution_method: ExecutionMethodResourceTarget::ExecutionMethod(
+                            CanisterMethod {
+                                canister_id: Principal::management_canister(),
+                                method_name: format!("method_{}", i),
+                            },
+                        ),
+                        validation_method: ValidationMethodResourceTarget::No,
```
mraszyk, line 341: Could we please also have a test having a validation method so that we test the ordering?

core/station/impl/src/services/external_canister.rs
```diff
@@ -517,196 +519,246 @@ impl ExternalCanisterService {
...
+                    // removes the specified policies if they are of the call type and exist for the external canister
+                    for policy_id in policy_ids_to_remove {
+                        if !current_policies.calls.contains(policy_id) {
+                            return Err(ExternalCanisterError::ValidationError {
+                                info: format!(
+                                    "The policy with id {} does not exist for the external canister.",
+                                    Uuid::from_bytes(*policy_id).hyphenated()
+                                ),
+                            })?;
+                        }
```
mraszyk, line 686: Will the subsequent removal not fail anyway (in other words, do we need this explicit check here)?

core/station/impl/src/services/external_canister.rs
```diff
@@ -718,32 +770,176 @@ impl ExternalCanisterService {
...
+                                                },
+                                            ),
+                                        validation_method: _,
+                                    },
+                                ),
+                            ) = &permission.resource
+                            {
+                                update_call_methods.contains(method_name)
+                            } else {
+                                false
```
mraszyk, line 873: we should never reach this branch, right?

core/station/impl/src/services/external_canister.rs
```diff
@@ -718,32 +770,176 @@ impl ExternalCanisterService {
...
+
+                        self.permission_service
+                            .edit_permission(EditPermissionOperationInput {
+                                auth_scope: Some(call.allow.auth_scope),
+                                users: Some(call.allow.users),
+                                user_groups: Some(call.allow.user_groups),
+                                resource: call_resource.clone(),
+                            })?;
+                    }
+                }
```
mraszyk, line 842: could we use a helper similarly to `maybe_mutate_canister_calls_request_policies`?

core/station/impl/src/services/external_canister.rs
```diff
@@ -718,32 +770,176 @@ impl ExternalCanisterService {
...
+                                                        canister_id: _,
+                                                        method_name,
+                                                    },
+                                                ),
+                                            validation_method: _,
+                                        },
+                                    ),
+                                ) = &permission.resource
+                                {
+                                    method_name == method
```
mraszyk, line 928: Could we check if the current method name is contained in methods instead of running a loop over methods?

core/station/impl/src/services/external_canister.rs
```diff
@@ -718,32 +770,176 @@ impl ExternalCanisterService {
...
+                            })?;
+                    }
+                }
+                ExternalCanisterChangeCallPermissionsInput::OverrideSpecifiedByExecutionMethods(
+                    calls,
+                ) => {
+                    let update_call_methods = calls.iter().fold(HashSet::new(), |mut acc, call| {
+                        acc.insert(call.execution_method.clone());
+                        acc
+                    });
```
mraszyk, line 849:
```suggestion
                    let update_call_methods: HashSet<_> = calls.iter().map(|call| call.execution_method).collect();
```

core/station/impl/src/services/external_canister.rs
keplervital: We do, because this validates that the user is not providing a request policy id that is not related to this external canister, otherwise they would be able to make this request and update any other policy in the system

core/station/impl/src/services/external_canister.rs
keplervital: correct, i've added a `matches!` here to be more concise. throwing an error would also be an option but maybe we could simply ignore it and not block the execution

mraszyk: sure, let's leave it with the matches!

core/station/impl/src/services/external_canister.rs
mraszyk: Oh, I see. Shouldn't this validation be performed by the authorization middleware?

core/station/impl/src/services/external_canister.rs
keplervital: just added `maybe_mutate_canister_calls_permissions`

core/station/impl/src/services/external_canister.rs
mraszyk: We agreed that it makes sense to have the check at both places: here and in the authorization middleware.

core/station/impl/src/services/external_canister.rs
```diff
@@ -517,196 +520,246 @@ impl ExternalCanisterService {
...
-        // add or update the `Call` policies
-        for policy in input.calls {
-            match policy.policy_id {
-                Some(policy_id) => {
-                    if !current_policies.contains(&policy_id) {
-                        print(format!(
-                            "Policy with id {} not found for external canister {}",
-                            Uuid::from_bytes(policy_id).hyphenated(),
-                            external_canister.canister_id.to_text()
-                        ));
```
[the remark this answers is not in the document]
mraszyk: And we should fail the request instead of continuing.

core/station/impl/src/services/external_canister.rs
```diff
@@ -517,196 +520,246 @@ impl ExternalCanisterService {
...
+                            external_canister,
+                            &self
+                                .request_policy_repository
+                                .find_external_canister_call_policies_by_execution_method(
+                                    &external_canister.canister_id,
+                                    &method,
+                                ),
+                            &calls,
+                        )?;
                     }
```
mraszyk, line 639: same here: we should add this back and ideally fail the request instead of continuing

core/station/impl/src/models/external_canister.rs
```diff
@@ -219,6 +231,176 @@ impl ModelValidator<ExternalCanisterError> for ExternalCanister {
...
+fn validate_change_policies_are_associated_with_target_canister(
+    canister_id: &Principal,
+    policy_ids: &[UUID],
+) -> ModelValidatorResult<ExternalCanisterValidationError> {
+    for policy_id in policy_ids {
+        // validates if the policy matches the expected change variant and canister_id or throws an error
+        assert_policy_is_associated_with_target_canister(policy_id, |policy| {
+            matches!(
+                policy.specifier,
+                RequestSpecifier::ChangeExternalCanister(ExternalCanisterId::Canister(id)) if id == *canister_id
```
mraszyk, line 272: How would an admin set a request policy for all canisters ExternalCanisterId::Any?

mraszyk: Using `RequestOperation::EditRequestPolicy`?

mraszyk (approved):

PR #369, opened 2024-10-02
feat(upgrader): set target canister's compute allocation if failed to stop

mraszyk:
This PR sets the target canister's compute allocation (temporarily) to 1 if the target canister failed to stop (and then retries to stop the target canister).


keplervital (approved):
LGTM

PR #371, opened 2024-10-03
ci(deploy): update deployment script to use artifacts

keplervital:


scripts/deploy.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -eEuo pipefail
 
+# Wether or not to reuse the artifacts that are already built
```
mraszyk, line 4:
```suggestion
# Whether or not to reuse the artifacts that are already built
```

scripts/deploy.sh
```diff
@@ -168,11 +186,13 @@ function deploy_control_panel() {
 
   echo "Building the control_panel wasm..."
```
mraszyk, line 187:
```suggestion
  echo "Preparing the control_panel wasm..."
```
just to be consistent with the above

scripts/deploy.sh
```diff
@@ -209,7 +229,33 @@ function deploy_app_wallet() {
...
+  canister_id_output=$(dfx canister id app_wallet --network $network 2>&1)
+  canister_id_exit_code=$?
+  set -e # Re-enable 'exit on error'
+
+  if [ $canister_id_exit_code -ne 0 ]; then
+    echo "Canister 'app_wallet' does not exist, creating and installing..."
+
+    BUILD_MODE=$network dfx deploy --network $network app_wallet --with-cycles 2000000000000 $([[ -n "$subnet_type" ]] && echo "--subnet-type $subnet_type")
+  else
+    echo "Deploying the app_wallet canister to the '$network' network is not yet supported automatically."
```
mraszyk, line 253: Why?

scripts/deploy.sh
```diff
@@ -209,7 +229,33 @@ function deploy_app_wallet() {
...
+    echo
+    echo "Please deploy the app_wallet canister manually using the following command:"
+    echo "icx-asset --pem YOUR_IDENTITY_PEM_PATH --replica TARGET_REPLICA_HTTP sync --no-delete $canister_id_output artifacts/wallet-dapp/dist"
```
mraszyk, line 256: Can't we just run this here instead of failing? If we do want to fail, maybe `exit 1` would be appropriate here?

keplervital: Actually I think we could require the identity pem path as an argument and that would enable us to run the command instead of just logging it, I’ll try that.

scripts/deploy.sh
keplervital: Because dfx is not exposing the —no-delete option of icx-asset

scripts/deploy.sh
```diff
@@ -106,15 +128,85 @@ function get_subnet_type() {
...
+    echo "$result"
+  fi
+}
+
+function set_identity_pem_path() {
+  local identity_name=$1
+  local identity_store_path=${DFX_DEFAULT_IDENTITY_STORE_PATH}
+
+  if [ -z "$IDENTITY_PEM_PATH" ]; then
+    local identity_pem_path="${identity_store_path}/identity_name/${identity_name}.pem"
```
mraszyk, line 160:
```suggestion
    local identity_pem_path="${identity_store_path}/${identity_name}/${identity_name}.pem"
```

scripts/deploy.sh
```diff
@@ -209,7 +303,33 @@ function deploy_app_wallet() {
...
+
+  mkdir -p ./artifacts/wallet-dapp/dist
+  tar -xvf ./artifacts/wallet-dapp/wallet-dapp.tar.gz -C ./artifacts/wallet-dapp/dist
+
+  set +e # Disable 'exit on error'
+  canister_id_output=$(dfx canister id app_wallet --network $network 2>&1)
+  canister_id_exit_code=$?
+  set -e # Re-enable 'exit on error'
+
+  if [ $canister_id_exit_code -ne 0 ]; then
```
mraszyk, line 322:
```suggestion
  if [ -z "$canister_id_output" ]; then
```

mraszyk (approved):

PR #378, opened 2024-10-08
feat(station): add ICRC-1 support

olaszakos:
This PR adds ICRC-1 support for the station.

Adding the default ICP asset to an account now generates 2 addresses for the account: AccountIdentifier and ICRC-1 Account. They reference the same account in the ICP ledger.

### DIFF
--- a/core/station/impl/src/factories/blockchains/internet_computer.rs
+++ b/core/station/impl/src/factories/blockchains/internet_computer.rs
@@ -431,52 +431,60 @@ impl BlockchainApi for InternetComputer {
     async fn balance(
         &self,
         asset: &Asset,
-        account_address: &AccountAddress,
+        account_addresses: &[AccountAddress],
     ) -> BlockchainApiResult<BigUint> {
-        match account_address.format {
-            AddressFormat::ICPAccountIdentifier => {
-                let balance = self
-                    .balance_of_account_identifier(
-                        asset,
-                        &ic_ledger_types::AccountIdentifier::from_hex(&account_address.address)
-                            .map_err(|error| BlockchainApiError::InvalidToAddress {
-                                address: account_address.address.clone(),
-                                error,
-                            })?,
-                    )
-                    .await?;
+        // all matching addresses should resolve to the same balance, so pick the first one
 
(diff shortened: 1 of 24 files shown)


core/station/impl/src/factories/blockchains/internet_computer.rs
```diff
@@ -137,17 +156,59 @@ impl InternetComputer {
...
+                info: format!("rejection_code: {:?}, err: {}", err.0, err.1),
+            })?
+            .0;
+
+        Ok(balance.0)
+    }
+
     pub fn transaction_fee(&self) -> u64 {
-        DEFAULT_FEE.e8s()
+        ic_ledger_types::DEFAULT_FEE.e8s()
```
keplervital, line 184: hardcoding this it assumes that we will only support ICP to be under this internet_computer implementation right? e.g. if OGY which also supports ICRC1 and ICP standard would want to use this they would not be able to if their fee would diverge

tests/integration/src/transfer_tests.rs
```diff
@@ -277,3 +287,301 @@ fn make_transfer_successful() {
...
+    let beneficiary_id = user_test_id(1);
+
+    // register user
+    let res: (ApiResult<MeResponse>,) =
+        update_candid_as(&env, canister_ids.station, WALLET_ADMIN_USER, "me", ()).unwrap();
+    let user_dto = res.0.unwrap().me;
+
+    let ledger_controller = Principal::from_slice(&[99; 29]);
+
+    let token_ledger_canister_id = deploy_icrc1_token(
```
keplervital, line 309: nit: this could pottentially be moved to a setup util method

core/station/impl/src/factories/blockchains/internet_computer.rs
olaszakos: good catch, will update this so it also fetches from the canister

tests/integration/src/transfer_tests.rs
olaszakos: I'd wait and see how often we need to have an icrc1 token deployed, my assumtion is not often.

PR #380, opened 2024-10-08
feat(dfx-orbit): support installing canisters with large WASM

mraszyk:
This PR fixes the computation of `module_checksum` for external canisters with large WASM in the station and adds support for large WASM in the dfx-orbit tool.


Sawchord (approved):
Looks good to me

keplervital (approved):
LGTM

PR #381, opened 2024-10-08
feat(station): add canister execution method and validation pair edit variant

keplervital:


core/station/api/spec.did
```diff
@@ -2315,6 +2327,18 @@ type ExternalCanisterPermissions = record {
...
+// The pair that is used to represent the execution and validation method.
+type CanisterExecutionAndValidationMethodPair = record {
+  // The validation method that is used to validate the request and
+  // render the argument.
+  validation_method : ValidationMethodResourceTarget;
+  // The method that the caller can call on the external canister.
+  //
+  // The `*` method name is used to represent that the caller can
+  // call any method on the canister.
+  execution_method : text;
```
mraszyk, line 2339: Could we use a typed struct instead of text to represent the case of Any (instead of using `*` for that)?

core/station/api/spec.did
```diff
@@ -1177,6 +1183,12 @@ type ListRequestsResult = variant {
 type GetRequestInput = record {
   // The request id to retrieve.
   request_id : UUID;
+  // Fill in all the additional info about the request operation, request types such as Call External Canisters will
+  // include the request argument and reply blobs, this can be a large amount of data and could potentially exceed the
+  // response size limit.
+  //
+  // By default, this field is set to `true`.
```
mraszyk, line 1190: What does "by default" mean? That none is unwrapped to true?

core/station/api/spec.did
```diff
@@ -774,6 +774,12 @@ type CallExternalCanisterOperation = record {
   // The reply blob produced by a successful call of the execution method,
   // i.e., when the request is `Completed`.
   execution_method_reply : opt blob;
+  // Optional details about the call operation, this field is not populated in list responses, only
+  // when using `get_request` to avoid going over the response size limit.
+  with_details : opt record {
```
mraszyk, line 779: why this additional layer instead of inlining arg : opt blob as a regular field of `CallExternalCanisterOperation`?

core/station/api/spec.did
```diff
@@ -774,6 +774,12 @@ type CallExternalCanisterOperation = record {
   // The reply blob produced by a successful call of the execution method,
   // i.e., when the request is `Completed`.
   execution_method_reply : opt blob;
+  // Optional details about the call operation, this field is not populated in list responses, only
+  // when using `get_request` to avoid going over the response size limit.
```
mraszyk, line 778:
```suggestion
  // when using `get_request` and setting `with_full_info` to `opt true` to avoid going over the response size limit.
```

core/station/api/spec.did
```diff
@@ -1177,6 +1183,12 @@ type ListRequestsResult = variant {
 type GetRequestInput = record {
   // The request id to retrieve.
   request_id : UUID;
+  // Fill in all the additional info about the request operation, request types such as Call External Canisters will
```
mraszyk, line 1186:
```suggestion
  // Fill in all the additional info about the request operation, request types such as `CallExternalCanisterOperation` will
```

core/station/api/spec.did
```diff
@@ -1177,6 +1183,12 @@ type ListRequestsResult = variant {
...
+  // include the request argument and reply blobs, this can be a large amount of data and could potentially exceed the
```
mraszyk, line 1187: > and reply blobs

does it mean that `execution_method_reply` in `CallExternalCanisterOperation` is now gonna be null if `with_full_info` is set to `opt false`?

core/station/api/spec.did
mraszyk: > this can be a large amount of data and could potentially exceed

shouldn't the default be false to avoid exceeding the limits?

core/station/impl/src/mappers/request.rs
```diff
@@ -57,6 +59,36 @@ impl Request {
                 .collect(),
         }
     }
+
+    pub fn to_dto_with_full_info(self) -> RequestDTO {
```
mraszyk, line 63: why not parameterize `fn to_dto` with a boolean if full info should be provided to avoid code duplication?

core/station/impl/src/repositories/indexes/request_policy_resource_index.rs
```diff
@@ -371,4 +403,58 @@ mod tests {
             assert_eq!(policies[0], expected_method_ids.pop().unwrap());
         }
     }
+
+    #[test]
+    fn test_find_external_canister_call_policies_by_execution_and_validation_method() {
```
mraszyk, line 408: Could we merge this test with the above one to avoid duplicating the common setup?

core/station/api/spec.did
```diff
@@ -2362,6 +2394,14 @@ type ExternalCanisterChangeCallRequestPoliciesInput = variant {
   RemoveByPolicyIds : vec UUID;
   // Override the request policies for the specified execution methods.
   OverrideSpecifiedByExecutionMethods : vec ExternalCanisterCallRequestPolicyRuleInput;
+  // Override the request policies for the specified execution and validation method pairs.
+  OverrideSpecifiedByExecutionValidationMethodPairs : vec record {
+    // The method configuration that is used to represent the execution and validation method pair.
+    method_configuration : CanisterExecutionAndValidationMethodPair;
```
mraszyk, line 2400: Why do we need to perform the grouping explicitly in the args as opposed to deriving the grouping from the flattened `vec ExternalCanisterCallRequestPolicyRuleInput` as for `OverrideSpecifiedByExecutionMethods`?

core/station/api/spec.did
mraszyk: this is irrelevant if we go for the approach from this [comment](https://github.com/dfinity/orbit/pull/381#discussion_r1793839137)

core/station/api/spec.did
mraszyk: and is this the reason why null defaults to `opt true` to avoid a breaking change (it is still a breaking change due to potentially exceeding size limits)

core/station/api/spec.did
keplervital: i would rather not, this is consistent to many other fields we already have for `execution_method` in `ExternalCanisterCallRequestPolicyRuleInput`, `ExternalCanisterCallRequestPolicyRule`, `ExternalCanisterCallPermission` and `ExternalCanisterCallerMethodsPrivileges`.

core/station/api/spec.did
keplervital: > What does "by default" mean? That none is unwrapped to true?

Correct, but actually i'll move this to `false` since this will be less risky for clients and if they set to true they will need to explicitely handle a possible fail condition.

core/station/api/spec.did
keplervital: The idea would be to make it very explicit that this is optional and is not a top level field of the main type, but maybe inlining arg would be good enough, wdyt?

core/station/api/spec.did
keplervital: as mentioned aboved, i'll update the default to be `false`, but execution_method_reply is still going to be returned, this will keep things backwards compatible.

core/station/impl/src/mappers/request.rs
keplervital: i created a `inner_to_dto` and kept the differentiation, mostly because i find it easier to ready being expressive then having an externally exposed boolean, wdyt?

keplervital: i created a `inner_to_dto` and kept the differentiation, mostly because i find it easier to ready being expressive then having an externally exposed boolean, wdyt?

core/station/api/spec.did
keplervital: Mostly because this way we can also remove in the same request, with the above we need to make 2 separate requests, e.g. in the UI with this you can enable the user to rename a method & validation pair and you would simply make one request that would remove the renamed method and add the new one.

core/station/api/spec.did
mraszyk: However, `*` is also a valid IC canister method name (I just checked by deploying a canister with such a method and calling it via `dfx`).

core/station/impl/src/repositories/indexes/request_policy_resource_index.rs
```diff
@@ -354,21 +386,39 @@ mod tests {
...
-            if i % 2 == 0 {
+            if i % 2 != 0 {
                 expected_method_ids.push([i; 16]);
             }
         }
 
         expected_method_ids.reverse();
 
-        for i in (0..20).step_by(2) {
+        for i in (1..20).step_by(2) {
```
mraszyk, line 396: Now we're losing coverage for cases of no validation method.

Also it'd be useful to test the case of two policies with the same execution method and (i) no validation, (ii) some validation while
- retrieving by execution method (expected to get them both)
- retrieving by execution and no validation method (expected to get one of them)
- retrieving by execution and some validation method (expected to get the other one of them)

core/station/api/spec.did
mraszyk: > rename a method & validation pair

can't this be risky in case there already exists a permission/policy for the new pair?

keplervital: I would say it’s up for the approver to check and block, on the UI if there’s another one we don’t let the user submit that request, this provides a better UX and helps avoid submitting an override request when there’s already another entry

core/station/impl/src/repositories/indexes/request_policy_resource_index.rs
keplervital: Good suggestion, I’ll extend the test tomorrow

core/station/api/spec.did
keplervital: Even if supported by a canister imo we should keep as is to avoid a possible migration, refactor at the moment and if someone has a need for it in the future we can revisit the decision.

core/station/api/spec.did
mraszyk: > Mostly because this way we can also remove in the same request

Doesn't that mean that we should group by execution method in the request in the other case, too, instead of providing a flattened list?

keplervital: correct, that would make it consistent, i'll make the change.

core/station/impl/src/repositories/indexes/request_policy_resource_index.rs
```diff
@@ -404,21 +402,125 @@ mod tests {
...
+                    &validation_method,
                 );
 
             assert_eq!(policies.len(), 1);
             assert_eq!(policies[0], expected_method_id);
         }
 
         assert!(expected_method_ids.is_empty());
+
+        // then insert another policy with the same execution method but different validation method
```
mraszyk, line 428: could we also add a policy with no validation (to test that no validation is sorted before some validation and take_while works as expected)?

core/station/api/src/external_canister.rs
```diff
@@ -67,11 +67,22 @@ pub struct ExternalCanisterCallPermissionMethodPairInput {
...
 }
 
+#[derive(CandidType, serde::Serialize, Deserialize, Debug, Clone)]
+pub struct ExternalCanisterCallPermissionExecMethodEntryInput {
+    pub allow: AllowDTO,
+    pub validation_method: ValidationMethodResourceTargetDTO,
+}
+
+#[derive(CandidType, serde::Serialize, Deserialize, Debug, Clone)]
+pub struct ExternalCanisterCallPermissionExecMethodListInput {
```
mraszyk, line 77: `List` can be probably dropped from the name, but this is a very minor thing.

core/station/impl/src/services/external_canister.rs
```diff
@@ -698,19 +698,21 @@ impl ExternalCanisterService {
...
+
+                        call.policies.iter().for_each(|policy| {
+                            entries.insert(ExternalCanisterCallRequestPolicyRuleInput {
+                                policy_id: policy.policy_id,
+                                rule: policy.rule.clone(),
+                                execution_method: call.execution_method.clone(),
+                                validation_method: policy.validation_method.clone(),
+                            });
+                        });
+                    });
```
mraszyk, line 715: Why are we not using the "same" implementation as for the variant `OverrideSpecifiedByExecutionValidationMethodPairs`?

mraszyk: I guess the difference is that this approach combines overrides with the same key: e.g., `[(method, [p1, p2]), (method, [p3])]` gets processed as `[(method, [p1, p2, p3])]` while the other approach overrides sequentially as `[(method, [p3])]`.

core/station/impl/src/controllers/request.rs
```diff
@@ -178,7 +178,10 @@ impl RequestController {
             .get_request_additional_info(&request, true)?;
 
         Ok(GetRequestResponse {
-            request: request.to_dto(),
+            request: match input.with_full_info {
+                None | Some(true) => request.to_dto_with_full_info(),
```
mraszyk, line 182: Didn't we want to switch the default?

keplervital: Yes, nice catch!

core/station/impl/src/repositories/indexes/request_policy_resource_index.rs
```diff
@@ -476,18 +492,29 @@ mod tests {
...
         // and find both policies by the execution method
         let policies = repository.find_external_canister_call_policies_by_execution_method(
             &Principal::management_canister(),
             "method_1",
         );
 
-        assert_eq!(policies.len(), 2);
+        assert_eq!(policies.len(), 3);
         assert!(policies.contains(&execution_method_policy_first_policy_id));
         assert!(policies.contains(&execution_method_policy_second_policy_id));
```
mraszyk, line 514: maybe also add
```
assert!(policies.contains(&execution_method_policy_third_policy_id));
```

core/station/impl/src/services/external_canister.rs
```diff
@@ -669,13 +669,6 @@ impl ExternalCanisterService {
                     )?,
                 ExternalCanisterChangeCallRequestPoliciesInput::OverrideSpecifiedByExecutionValidationMethodPairs(calls) => {
                     for call in calls {
-                        let initial_policy_ids = self.request_policy_repository
```
mraszyk: Is there a reason to inline the let binding?

keplervital: mostly to make `OverrideSpecifiedByExecutionValidationMethodPairs` and `OverrideSpecifiedByExecutionMethods` consistent, no strong preference through

core/station/impl/src/services/external_canister.rs
mraszyk: I guess in both cases, we assume that keys won't be repeated, but I'm not sure we're rejecting such inputs as invalid.

keplervital: in OverrideSpecifiedByExecutionMethods, if it's repeated then it will be aggregated and the repeated values would be removed, but for `OverrideSpecifiedByExecutionValidationMethodPairs` we don't do that, but i think we should, i'm adding that change

mraszyk (approved):

PR #385, opened 2024-10-11
feat(wallet): add external canister method call ui

keplervital:
Adds the necessary missing components to surface the method call to an external canister, this also include the configuration of the approvals, policies, etc... for those external method calls.

At the moment the argument and reply constructions is not very helpful because we need to provide hex encoded argument and we get back a `UInt8Array` back as a response, as a follow up PR i'll give it a try to use candid parser to fetch the public metadata of the canister and have a better DX for that.

I'll mark the PR as ready once the following is completed:

- [x] Test coverage

**Views:**

Screenshot 2024-10-11 at 20 34 15
Screenshot 2024-10-11 at 20 32 53
Screenshot 2024-10-11 at 20 32 30
Screenshot 2024-10-11 at 20 32 09
Screenshot 2024-10-11 at 20 31 59


olaszakos:
Enable external canister - List for admins by default?

olaszakos:
Maybe add some info that the validation method must return with `{ Ok : text; Err : text }`?

apps/wallet/src/utils/helper.utils.ts
```diff
@@ -238,6 +239,14 @@ export const toArrayBuffer = (input: Uint8Array | number[]): ArrayBuffer => {
   return input instanceof Uint8Array ? input.buffer : new Uint8Array(input).buffer;
 };
 
+export const toUint8Array = (input: ArrayBuffer | number[] | Uint8Array): Uint8Array => {
+  if (input instanceof Uint8Array) {
+    return input;
+  }
+
+  return input instanceof ArrayBuffer ? new Uint8Array(input) : new Uint8Array(input);
```
olaszakos, line 247: Redundant check?

olaszakos (approved):

PR #389, opened 2024-10-18
chore(upgrader): handle unstoppable station in disaster recovery

mraszyk:
This PR handles the case of an unstoppable station during disaster recovery:
- the station canister must always be stopped before installing code (otherwise, it is unsafe to install new code due to open call contexts that might trigger executing unrelated callbacks on the new wasm) and if the station canister cannot be stopped (the ICP has a timeout of 5 mins for stopping), then disaster recovery fails
- there's a new Boolean option `force` of disaster recovery to force disaster recovery of an unstoppable canister by taking a snapshot, uninstalling the canister (marking all open call contexts as deleted), restoring the snapshot, and resuming with installing new code (it is safe to not actually stop the canister in this case because all open call contexts are marked as deleted when uninstalling code); this makes sense for an unstoppable Orbit station because a station has all its data in stable memory that can be fully restored from a snapshot


tests/integration/src/disaster_recovery_tests.rs
```diff
@@ -997,3 +1009,182 @@ fn test_disaster_recovery_failing() {
...
+            snapshots.is_empty()
+                || (snapshots.len() == 1 && current_snapshots[0].id != snapshots[0].id)
+        );
+        snapshots = current_snapshots;
+
+        // check new station name set during disaster recovery
+        let system_info = get_system_info(&env, WALLET_ADMIN_USER, canister_ids.station);
+        assert_eq!(system_info.name, new_name);
+
+        // the call request will be "Processing" forever since we deleted its call context during disaster recovery
```
keplervital, line 1175: not related to this PR, but we should ideally have something in place to perform some cleanup on requests that are forever `processing` since they are actually not processing, it's just dangling state

core/upgrader/impl/src/services/disaster_recovery.rs
```diff
@@ -251,63 +254,75 @@ impl DisasterRecoveryService {
...
+                })
+                .await
+                .map_err(|(_code, msg)| msg)?;
+                load_canister_snapshot(LoadCanisterSnapshotArgs {
+                    canister_id: station_canister_id,
+                    snapshot_id: snapshot.id,
+                    sender_canister_version: None,
+                })
+                .await
+                .map_err(|(_code, msg)| msg)?;
```
keplervital, line 306: if this would fail for some reason (e.g. because maybe the target canister would not have enough cycles), then the disaster recovery committee would need to call this endpoint again, however, because we replace that snapshot, then the second try would take a snapshot of an empty canister because we would have already uninstalled it.

To account for such cases, maybe we should store a flag in the upgrader that makes it reuse the same snapshot id on a retry, wdyt?

mraszyk: Great catch! I'd suggest the following:
- support taking and restoring snapshots manually in disaster recovery
- never take a snapshot automatically: in the force mode a manually created snapshot must be specified, i.e., `force: bool` becomes `force: Option<SnapshotId>`

keplervital: Both look like a safer approach, and in the case of `force: Option<SnapshotId>`, if provided as `None` we would then create the snapshot? as it is the behaviour for `force=true`

mraszyk: `force: None` would correspond to `force: false` and `force: Some(snapshot_id)` would correspond to `force: true`

PR #392, opened 2024-10-21
chore(orbit-essentials): use asset canister native chunking for large WASM

mraszyk:
This PR uses the asset canister native chunking for large WASM: the extra chunks are uploaded as a single (chunked) asset as opposed to uploading each chunk as a separate asset.

### DIFF
--- a/libs/orbit-essentials/src/install_chunked_code.rs
+++ b/libs/orbit-essentials/src/install_chunked_code.rs
@@ -7,12 +7,22 @@ use ic_cdk::api::management_canister::main::{
 use ic_cdk::call;
 use sha2::{Digest, Sha256};
 
-// asset canister argument types
+const MAX_ICP_CHUNK_LEN: usize = 1 << 20;
+
+// asset canister types
 
 #[derive(CandidType)]
 struct GetArg {
     pub key: String,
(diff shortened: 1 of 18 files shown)


libs/orbit-essentials/src/install_chunked_code.rs
```diff
@@ -26,27 +36,77 @@ struct EncodedAsset {
...
+            key: extra_chunks_key.clone(),
+            accept_encodings: vec!["identity".to_string()],
+        },),
+    )
+    .await
+    .map_err(|(_, err)| format!("failed to fetch asset: {err}"))?
+    .0;
+    let mut res = asset.content;
+    let mut idx = 1_u64;
+    while res.len() < asset.total_length {
```
keplervital, line 63: let's prevent an asset canister from creating an infinite loop, ideally we can allow max number of iterations and same for max_length, if > then what the canister accepts (iirc 100mib) then we can directly fail this even before the loop.

mraszyk: Fixed: https://github.com/dfinity/orbit/pull/392/commits/c15fc12d35bab66069f653ad59e0dadbf770f3d9

PR #394, opened 2024-10-21
feat(station): Auto-approve requests from controllers when they are created

anodar-df:


anodar-df:
As discussed offline, we're going with different approach. Closing this.

PR #395, opened 2024-10-21
feat(dfx-orbit): Enable controller management through Orbit

Sawchord:
This PR enables the update_settings feature in `dfx-orbit`.

Currently, it only supports adding and removing controllers.
Test is included


tests/integration/src/dfx_orbit/manage_settings.rs
```diff
@@ -0,0 +1,159 @@
...
+                    user_groups: vec![],
+                    users: vec![],
+                },
+            },
+            request_policies: ExternalCanisterRequestPoliciesCreateInput {
+                change: vec![],
+                calls: vec![],
+            },
+        }),
+    );
```
mraszyk, line 71: let's please wrap this into a helper

tests/integration/src/dfx_orbit/manage_settings.rs
```diff
@@ -0,0 +1,159 @@
...
+            action: RequestArgsActions::Canister(RequestCanisterArgs {
+                action: RequestCanisterActionArgs::UpdateSettings(add_controller_args.clone()),
+            }),
+        }
+        .into_request(&dfx_orbit)
+        .await
+        .unwrap();
+        dfx_orbit.station.request(request.clone()).await.unwrap();
+
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
mraszyk, line 116: let's properly wait until the request has been completed

tests/integration/src/dfx_orbit/manage_settings.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        let controllers = dfx_orbit.get_controllers(canister_id).await.unwrap();
+        assert_eq!(controllers, vec![user_test_id(1), canister_ids.station]);
```
mraszyk, line 119: let's not check for list equality since the order of controllers could change, instead  we should check that the list has a length of two and contains both elements

tests/integration/src/dfx_orbit/util.rs
```diff
@@ -61,6 +61,17 @@ pub(super) fn permit_change_operation(env: &PocketIc, canister_ids: &CanisterIds
     permit_operation(env, canister_ids, resource);
 }
 
+/// Allow anyone to read request list
+pub(super) fn permit_list_reads(env: &PocketIc, canister_ids: &CanisterIds) {
+    let add_permission = RequestOperationInput::EditPermission(EditPermissionOperationInput {
```
mraszyk, line 66:
```suggestion
    let edit_permission = RequestOperationInput::EditPermission(EditPermissionOperationInput {
```

tools/dfx-orbit/src/lib.rs
```diff
@@ -150,9 +150,18 @@ impl DfxOrbit {
...
-            bail!("Expected an array as result from controllers endpoint")
+
+        let array = match value {
+            ciborium::Value::Array(array) => array,
+            ciborium::Value::Tag(_, inner) => {
+                let Ok(array) = inner.into_array() else {
+                    bail!("Expected an array as result from controllers endpoint");
+                };
+                array
+            }
```
mraszyk, line 161: Why was this not needed before?

tools/dfx-orbit/src/canister.rs
```diff
@@ -31,7 +31,7 @@ pub enum RequestCanisterActionArgs {
     /// Request to call a canister method
     Call(RequestCanisterCallArgs),
     // Update a canister's settings (i.e its controller, compute allocation, or memory allocation.)
-    //UpdateSettings(RequestCanisterUpdateSettingsArgs),
```
mraszyk, line 34: Why was this commented out before?

tools/dfx-orbit/src/lib.rs
Sawchord: There was a breaking change in the station, and we could not use this function. Therefore I did not surface the call to the  CLI level, and also I didn't have time to implement a test before.

I did both of these things now, and found out that there was a bug in the cbor parsing, while testing. So this was needed all along.

tools/dfx-orbit/src/canister.rs
Sawchord: As written above, I did not surface this code to the CLI on purpose, because Kepler told me that there would be a breaking change.

tests/integration/src/dfx_orbit/manage_settings.rs
Sawchord: I added a small function that sorts both vecs before comparing them.

tests/integration/src/dfx_orbit/manage_settings.rs
Sawchord: Implemented a function that pools the request until it has completed, fails if there is a failed state and also can time out

tests/integration/src/dfx_orbit/manage_settings.rs
Sawchord: After talking to @keplervital , it turns out that it is possible to remove the requirement for the registration altogether. This was implemented in #396. It allowed me to remove the code entirely.

mraszyk (approved):

PR #399, opened 2024-10-22
fix(dfx-orbit): uploading chunks to asset canister

mraszyk:
This PR bumps ic-agent, ic-utils, and dfinity/sdk dependencies to their more recent versions to fix a bug in dfx-orbit: chunks were not uploaded to the asset canister in the right order. The fix has been tested manually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,16 +41,16 @@ cap-std = "3.1.0"
 ciborium = "0.2.2"
 clap = { version = "4.5.7", features = ["derive"] }
 dateparser = "0.2"
-dfx-core = { git = "https://github.com/dfinity/sdk.git", rev = "bb5f8b58afa94b1950f5e1a750e0491457ad88d1" }
+dfx-core = { git = "https://github.com/dfinity/sdk.git", rev = "705c34689a1e5cb2c8421404f3c120fff93e4612" }
 flate2 = "1.0"
 convert_case = "0.6"
 futures = "0.3"
 getrandom = { version = "0.2", features = ["custom"] }
 hex = "0.4"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -41,24 +41,24 @@ cap-std = "3.1.0"
...
+dfx-core = { git = "https://github.com/dfinity/sdk.git", rev = "705c34689a1e5cb2c8421404f3c120fff93e4612" }
 flate2 = "1.0"
 convert_case = "0.6"
 futures = "0.3"
 getrandom = { version = "0.2", features = ["custom"] }
 hex = "0.4"
-ic-agent = { git = "https://github.com/dfinity/agent-rs.git", rev = "be929fd7967249c879f48f2f494cbfc5805a7d98" }
-ic-asset = { git = "https://github.com/dfinity/sdk.git", rev = "bb5f8b58afa94b1950f5e1a750e0491457ad88d1" }
+ic-agent = "0.38"
+ic-asset = { git = "https://github.com/dfinity/sdk.git", rev = "705c34689a1e5cb2c8421404f3c120fff93e4612" }
```
Sawchord, line 51: This version does not have the reproducibility fix merged yet and would reintroduce a bug. I just merged the fix into SDK main, and was about to change the version accordingly.

Sawchord: I would close this PR in favor of #400

mraszyk: bumped to d65717bd6d0c172247c37dd23395c9fb13b2ba20 - is this enough?

Sawchord: Ok, yeah, now they are identical 😅  I guess we can go ahead with this one

PR #406, opened 2024-10-25
feat(dfx-orbit): Implement station file option

Sawchord:
This PR enables to specify a  `--station-file`, to load the station configuration from, rather than using station files  stored via `dfx-orbit station add`.

The rationale is, that this makes using `dfx-orbit` easier from CI, and to distribute station files with repositories.


keplervital (approved):
LGTM

PR #411, opened 2024-11-01
chore(ci): support large wasm in orbit script

mraszyk:
This PR adds support for large wasm to the orbit local deployment script. This PR has been tested manually with a large WASM exceeding the ingress message limit.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -63,6 +63,12 @@
       "type": "assets",
       "source": ["apps/wallet/dist/"],
       "build": "pnpm --filter 'wallet-dapp' build"
+    },
+    "wasm_chunk_store": {
+      "type": "custom",
+      "candid": "./wasms/assetstorage.did",
+      "build": "./scripts/download-asset-canister-wasm.sh",
+      "wasm": "./wasms/assetstorage.wasm.gz"
     }
   },
   "defaults": {
--- a/dfx.json
+++ b/dfx.json
@@ -65,10 +65,9 @@
       "build": "pnpm --filter 'wallet-dapp' build"
     },
     "wasm_chunk_store": {
-      "type": "custom",
-      "candid": "./wasms/assetstorage.did",
-      "build": "./scripts/download-asset-canister-wasm.sh",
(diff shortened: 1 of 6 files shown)


dfx.json
```diff
@@ -63,6 +63,12 @@
       "type": "assets",
       "source": ["apps/wallet/dist/"],
       "build": "pnpm --filter 'wallet-dapp' build"
+    },
+    "wasm_chunk_store": {
+      "type": "custom",
```
olaszakos, line 68: can this not be something simple like

```
"wasm_chunk_store": {
      "type": "assets",
      "source": []
    },
```

not sure if this is correct dfx.json, but this would let you not need to download the wasm/candid.

olaszakos (approved):
looks good, `orbit --init` worked locally

dfx.json
mraszyk: Not sure why, but the frontend fails to load if I do this change.

mraszyk: It looks like a dfx bug to me - this dfx.json file in a sample project created by dfx new
```
{
  "canisters": {
    "test_backend": {
      "main": "src/test_backend/main.mo",
      "type": "motoko"
    },
    "test_frontend": {
      "dependencies": [
        "test_backend"
      ],
      "source": [
        "src/test_frontend/dist"
      ],
      "type": "assets",
      "workspace": "test_frontend"
    },
    "test": {
      "type": "assets",
      "source": []
    }
  },
  "defaults": {
    "build": {
      "args": "",
      "packtool": ""
    }
  },
  "output_env_file": ".env",
  "version": 1
}
```
results in `dfx deploy --no-wallet` being flaky which is a sign of a bug in my opinion.

mraszyk: not sure if that bug is related to our issue though

mraszyk: Adding `"build": ""` fixed it for me. Could you please confirm the local deployment works for you on my recent commit, @olaszakos ?

olaszakos: I can confirm that local deployment works now.

PR #420, opened 2024-11-18
fix(station): hardening validation

olaszakos:


core/station/impl/src/core/validation.rs
```diff
@@ -199,11 +195,23 @@ impl EnsureExternalCanister {
...
-        // todo: add a mutable denylist for external canisters
+        // Check if the target canister is a ledger canister of an asset.
+        let asset_canister_ids = ASSET_REPOSITORY
+            .list()
+            .iter()
+            .filter_map(|asset| {
+                asset
+                    .metadata
+                    .get(TokenStandard::METADATA_KEY_LEDGER_CANISTER_ID)
+            })
```
keplervital, line 206: nit: alternatively you could simply return early if found instead of collecting the list

keplervital (approved):

core/station/impl/src/core/validation.rs
olaszakos: good point, updated

keplervital (approved):

PR #421, opened 2024-11-18
refactor(control-panel): remove monitoring in control panel

jedna:


jedna:
We will continue to subsidize stations for now with more restrictions on the funding amount. More details: https://github.com/dfinity/orbit/pull/422

PR #424, opened 2024-11-19
feat(station): add expiration dt during request creation

keplervital:
The requests have an expiration date, however, up until now that was forced to 30 days in the Orbit Station and was not exposed to the API, this change enables request creators to set the expiration date.

### DIFF
--- a/core/station/impl/src/factories/requests/add_address_book_entry.rs
+++ b/core/station/impl/src/factories/requests/add_address_book_entry.rs
@@ -1,7 +1,7 @@
 use super::{Create, Execute, RequestExecuteStage};
 use crate::{
     errors::{RequestError, RequestExecuteError},
-    models::{AddAddressBookEntryOperation, Request, RequestExecutionPlan, RequestOperation},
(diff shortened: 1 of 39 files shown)


core/station/impl/src/factories/requests/add_address_book_entry.rs
```diff
@@ -18,22 +18,15 @@ impl Create<station_api::AddAddressBookEntryOperationInput> for AddAddressBookEn
...
             }),
-            input
-                .execution_plan
-                .map(Into::into)
-                .unwrap_or(RequestExecutionPlan::Immediate),
-            input
-                .title
-                .unwrap_or_else(|| "Address book entry creation".to_string()),
-            input.summary,
+            "Add address book entry".to_string(),
```
olaszakos, line 29: nit: create/add are inconsistent

olaszakos (approved):
LGTM

core/station/impl/src/factories/requests/add_address_book_entry.rs
keplervital: let's use `add` then

PR #429, opened 2024-11-20
feat(station): external canister snapshots

mraszyk:


core/station/api/spec.did
```diff
@@ -2646,6 +2688,21 @@ type GetExternalCanisterFiltersResult = variant {
   Err : Error;
 };
 
+type CanisterSnapshotsInput = record {
+  canister_id : principal;
+};
+
+type CanisterSnapshotsResponse = vec record {
+  id : blob;
+  taken_at_timestamp : nat64;
```
keplervital, line 2697: for dates we should use a string with `TimestampRFC3339`

mraszyk: I've changed the API types [here](https://github.com/dfinity/orbit/pull/429/commits/38206bd0aac66346f59ddd5f424962fedd52bb7a), trapping if conversion fails.

keplervital (approved):

PR #431, opened 2024-11-21
feat(wallet): stop monitoring for external canister

jedna:


apps/wallet/src/locales/cs.locale.ts
```diff
@@ -0,0 +1,1037 @@
+export default {
```
keplervital, line 1: let's not add `cs` just yet 🙂

PR #432, opened 2024-11-21
fix(station): Make DR sync + logging more efficient

olaszakos:
Problem:

1. The station syncs all assets and accounts to the upgrader on any account and asset change. The accounts change on each balance query since the new balance is saved back to the repository. This results in very frequent unnecessary syncs.
2. The upgrader log size is growing rapidly. Trimming is necessary, but it cannot be trimmed due to the stable Log structure only supporting append operation. 

Task:
1. [x] only sync assets and accounts when the change is relevant - ie. any field of the DR version of the entries change
2. [x] switch to a stable BTreeMap to keep the logs and limit the log size to 25k items

Also bundled a small fix for a missing asset foreign key check.


core/station/impl/src/services/disaster_recovery.rs
```diff
@@ -201,13 +176,87 @@ pub fn disaster_recovery_observes_remove_user(observer: &mut Observer<User>) {
...
+    T: SyncEq,
+{
+    observer.add_listener(Box::new(|sync_cmp| {
+        if !SYSTEM_SERVICE.is_healthy() {
+            // Skip syncing during system init
+            return;
+        }
+
+        if sync_cmp.sync_eq() {
+            // Skip syncing if the account hasn't changed
```
keplervital, line 251:
```suggestion
            // Skip syncing if the account or asset hasn't changed
```

keplervital (approved):
LGTM

PR #433, opened 2024-11-21
feat(enable): wallet ui accepts wasm chunk store from registry

keplervital:
Updates the control panel registry CLI script to operate with the large wasms, as well as, the frontend to correctly work if the control panel returns the large wasm information.

### DIFF
--- a/apps/wallet/src/components/system-upgrade/RegistryUpdateMode.vue
+++ b/apps/wallet/src/components/system-upgrade/RegistryUpdateMode.vue
@@ -134,7 +134,8 @@ const prepareUpdateWithWasmModule = async (): Promise<void> => {
 
     modelValue.value = {
       ...modelValue.value,
-      wasmModule: toArrayBuffer(artifact.artifact),
+      wasmModule: artifact.artifact.length ? toArrayBuffer(artifact.artifact) : undefined,
+      wasmModuleExtraChunks: registryEntry.value.WasmModule.module_extra_chunks?.[0],
       wasmInitArg: undefined,
     };
 
--- a/cli/src/registry/publish.ts
+++ b/cli/src/registry/publish.ts
@@ -189,7 +189,7 @@ command.action(async options => {
 
     const moduleStoredFileName =
       `${entry.name}_${entry.version}_${moduleHashHex.substring(0, 8)}`.replace(
-        /[^a-zA-Z0-9_]/g,
+        /[^a-zA-Z0-9_.]/g,
         '',
       );
     const tempFilePath = join(tmpdir(), moduleStoredFileName);
--- a/cli/src/registry/publish.ts
+++ b/cli/src/registry/publish.ts
@@ -188,8 +188,8 @@ command.action(async options => {
     const moduleUint8Array = new Uint8Array(moduleHashBuffer);
 
     const moduleStoredFileName =
-      `${entry.name}_${entry.version}_${moduleHashHex.substring(0, 8)}`.replace(
-        /[^a-zA-Z0-9_.]/g,
+      `${entry.name}_v${entry.version}_sha256-${moduleHashHex.substring(0, 8)}`.replace(
+        /[^a-zA-Z0-9_.-]/g,
         '',
       );
     const tempFilePath = join(tmpdir(), moduleStoredFileName);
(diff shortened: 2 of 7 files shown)


cli/src/utils.ts
```diff
@@ -73,3 +80,54 @@ export const assertReplicaIsHealthy = async (network: string): Promise<void> =>
...
+  }
+
+  if (dfxFile.networks[network].providers && dfxFile.networks[network].providers.length > 0) {
+    return dfxFile.networks[network].providers[0];
+  }
+
+  if (dfxFile.networks[network].bind) {
+    return dfxFile.networks[network].bind.startsWith('http')
+      ? dfxFile.networks[network].bind
+      : `http://${dfxFile.networks[network].bind}`;
```
mraszyk, line 109: is this http:// a safe choice?

cli/src/registry/registry.core.ts
```diff
@@ -1,11 +1,19 @@
...
 
 export enum Application {
   Station = 'station',
   Upgrader = 'upgrader',
 }
 
+export const getWasmChunkStoreId = (network: string = 'local'): Principal => {
+  const maybeCanisterId = execSync(`dfx canister id wasm_chunk_store --network ${network}`);
+
+  return Principal.fromText(maybeCanisterId.toString().split('\n').join(''));
```
mraszyk, line 14: why split and join?

cli/src/registry/publish.ts
```diff
@@ -170,6 +183,26 @@ command.action(async options => {
...
     )?.id;
 
+    const moduleHashBuffer = createHash('sha256').update(entry.module).digest();
+    const moduleHashHex = moduleHashBuffer.toString('hex');
+    const moduleUint8Array = new Uint8Array(moduleHashBuffer);
+
+    const moduleStoredFileName =
+      `${entry.name}_${entry.version}_${moduleHashHex.substring(0, 8)}`.replace(
+        /[^a-zA-Z0-9_]/g,
+        '',
```
mraszyk, line 193: why this replace needed?

apps/wallet/src/components/system-upgrade/RegistryUpdateMode.vue
```diff
@@ -134,7 +134,8 @@ const prepareUpdateWithWasmModule = async (): Promise<void> => {
 
     modelValue.value = {
       ...modelValue.value,
-      wasmModule: toArrayBuffer(artifact.artifact),
+      wasmModule: artifact.artifact.length ? toArrayBuffer(artifact.artifact) : undefined,
```
mraszyk, line 137: what does this ternary operator do (in particular, why undefined)?

cli/src/utils.ts
keplervital: Yes, bind is only used for localhost ifaik

mraszyk: > bind is only used for localhost ifaik

how about failing the script if this assumption is wrong?

apps/wallet/src/components/system-upgrade/RegistryUpdateMode.vue
keplervital: In JS properties start as `undefined`, in this case if there is an artifact we add it, otherwise, we set as if the wasmModule was not set, hence the undefined.

mraszyk: I see, thank you for the explanation!

cli/src/registry/registry.core.ts
keplervital: Actually i can just use `trim` since in JS that also removes new lines.

cli/src/registry/publish.ts
keplervital: i was not sure if the asset canister would have restrictions for the filenames

mraszyk: Where does the name come from? I mean: why would it have nonstandard characters at the first place?

keplervital: the non standard characters come from the package name, which we publish to the registry as `@orbit/station` or `@orbit/upgrader`, and that would also break the file name due to `/`

mraszyk (approved):

PR #437, opened 2024-11-22
feat(station): prune external canister request

mraszyk:
This PR adds a new request type to prune a resource (a particular snapshot, the entire chunk store, or the entire state - heap and stable memory) of an external canister.


core/station/api/spec.did
```diff
@@ -847,6 +847,21 @@ type RestoreExternalCanisterOperation = record {
   input : RestoreExternalCanisterOperationInput;
 };
 
+type PruneExternalCanisterOperationInput = record {
+  // The canister to prune.
+  canister_id : principal;
+  // The resource to prune.
+  prune : variant {
+    chunk_store;
+    snapshot : blob;
```
keplervital, line 856: maybe we should surface a hex encoded version of this snapshot through the api, wdyt?

keplervital (approved):

core/station/api/spec.did
mraszyk: I've changed this [here](https://github.com/dfinity/orbit/pull/437/commits/e7fc7274c418f7506ad309172d9d6e52057d9187), trapping if the conversion from hex fails.

keplervital (approved):

PR #440, opened 2024-11-25
perf(station): make concurrent balance queries more efficient

olaszakos:
At the moment concurrent calls to fetch the balances of the assets of an account result in all of them fetching the balances with inter canister calls.

If an account asset balance is getting queried concurrent calls should not result in additional balance queries, but those should return the last balance value.

### DIFF
--- a/core/station/impl/src/models/account.rs
+++ b/core/station/impl/src/models/account.rs
@@ -1,5 +1,6 @@
 use super::{AccountBalance, AssetId};
 use crate::core::validation::{EnsureAsset, EnsureIdExists};
+use crate::core::ACCOUNT_BALANCE_FRESHNESS_IN_MS;
 use crate::errors::{AccountError, RecordValidationError};
 use crate::models::Metadata;
 use crate::repositories::request_policy::REQUEST_POLICY_REPOSITORY;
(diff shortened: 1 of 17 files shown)


core/station/impl/src/models/account.rs
```diff
@@ -249,6 +250,35 @@ impl Account {
...
+            BalanceQueryState::Fresh => write!(f, "fresh"),
+        }
+    }
+}
+
+impl From<&AccountBalance> for BalanceQueryState {
+    fn from(balance: &AccountBalance) -> Self {
+        let balance_age_ms = crate::core::ic_cdk::api::time()
+            .saturating_sub(balance.last_modification_timestamp)
+            / 1000;
```
keplervital, line 273: time get's back the time in nanoseconds, to get the miliseconds you need to divide by `1_000_000`

core/station/impl/src/services/account.rs
```diff
@@ -510,9 +527,34 @@ impl AccountService {
...
+                        let balance_update_guard_key = BalanceFetchGuardKey {
+                            account_id: account.id,
+                            asset_id: asset.id,
+                        };
+
+                        let _lock = BALANCE_FETCH_GUARD_STATE.with(|state| {
+                            CallerGuard::new(
+                                state.clone(),
+                                balance_update_guard_key,
+                                Some(time() + Duration::from_secs(60).as_nanos() as u64),
```
keplervital, line 541: maybe better to increase this to `5min` in case the subnet would be with high load

keplervital (approved):
LGTM

PR #445, opened 2024-11-27
feat(wallet): Cycle Obtain configuration for external canister

jedna:


apps/wallet/src/components/settings/obtain-cycle/MintFromNativeToken.vue
```diff
@@ -47,13 +48,19 @@ const model = computed({
...
 
-const accountList = computed(() => {
-  const accounts = autocomplete.results.value.map(group => ({
-    title: group.name,
-    value: group.id,
-  }));
+const station = useStationStore();
+
+const [icpAsset] = station.configuration.details.supported_assets.filter(
+  asset => asset.symbol === 'ICP',
```
keplervital, line 54: we need to check the ledger id instead of the symbol, the ledger_id should be `ryjl3-tyaaa-aaaaa-aaaba-cai`

jedna: updated

keplervital (approved):

PR #460, opened 2024-12-06
chore(control-panel,station): always use CMC for canister creation

mraszyk:
This PR updates the helper function for canister creation to always use CMC for canister creation: this way, stations created by the control panel and external canisters created by the station without specifying a subnet explicitly will be created at a (pseudo-)random application subnet instead of always creating those canisters on the same subnet. (The upgrader is still always created on the same subnet as its station.) This is supposed to achieve better load balancing. If external canisters need to be collocated, a subnet should be specified explicitly.


keplervital (approved):
LGTM

PR #461, opened 2024-12-13
chore(upgrader): introduce disaster recovery request operation

mraszyk:
This PR introduces a disaster recovery request operation type in the upgrader canister (to later support more operation types beyond the existing install code operation type).


core/upgrader/impl/src/controllers/mod.rs
```diff
@@ -3,3 +3,18 @@ mod logs;
...
+
+#[post_upgrade]
+fn post_upgrade() {
+    if !DISASTER_RECOVERY_SERVICE
+        .storage
+        .get()
+        .recovery_requests
+        .is_empty()
+    {
+        ic_cdk::trap("upgrader cannot be upgraded due to pending disaster recovery requests")
```
olaszakos, line 18: How do you recover from this state? The only place where `recovery_requests` is mutated is in `evaluate_requests`, but that only does it if the recovery rule is met. It does call `retain` on the requests to clean up the expired ones, but does not actually save it unless the requests meet evaluation.

Also at this point the incompatible stable memory is deserialized, no? So in case the vec is not empty, this has already trapped, no?

mraszyk: > How do you recover from this state?

The expectation is that the upgrader canister is not itself upgraded if there are pending disaster recoveries.

> Also at this point the incompatible stable memory is deserialized, no? So in case the vec is not empty, this has already trapped, no?

That's true, the error message indeed refers to failed deserialization.

olaszakos: > The expectation is that the upgrader canister is not itself upgraded if there are pending disaster recoveries.
But it can happen that a request is there for some reason, and then the upgrader can no longer be upgraded until the next recovery request passes.

> That's true, the error message indeed refers to failed deserialization.
What I mean is that maybe the `is_empty()` call already traps and the subsequent trap is ineffective as execution never gets there. I don't know if `is_empty` already deserializes something incompatible, that depends on the internals of the stable memory library. Maybe adopt the custom serde deserializer-based migration from the station? And then the requests can safely be migrated.

olaszakos (approved):

PR #465, opened 2024-12-18
fix(ci): install CMC in Orbit local deployment

mraszyk:
This PR installs CMC ([proposal](https://dashboard.internetcomputer.org/proposal/134500)) in Orbit local deployments since CMC is required to create canisters (stations and external canisters). Additionally, this PR:
- updates README to use PocketIC as the local IC backend (to retrieve its topology at an HTTP endpoint not served by the replica) and stop it using `dfx stop` rather than by CTRL^C (due to an issue in dfx)
- updates dfx version (to include recent PocketIC serving its topology at an HTTP endpoint)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -152,7 +152,7 @@ jobs:
       - name: 'Install dfx'
         uses: dfinity/setup-dfx@main
       - name: 'Start local replica'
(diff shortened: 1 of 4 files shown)


jedna (approved):
Tested, works. Thanks!

PR #466, opened 2024-12-19
chore(orbit-essentials): clear ICP chunk store after code install

mraszyk:


libs/orbit-essentials/src/install_chunked_code.rs
```diff
@@ -181,3 +183,31 @@ pub async fn install_chunked_code(
...
+        module,
+        module_extra_chunks,
+        arg,
+    )
+    .await;
+    if has_extra_chunks {
+        // clear the ICP chunk store of the target canister to reduce memory footprint
+        // and ignore any errors here so that a successful code install is not reported
+        // as failed only because of a failure to clear the ICP chunk store
+        let _ = mgmt::clear_chunk_store(ClearChunkStoreArgument {
```
keplervital, line 207: any reason to create a separate function for this call instead of adding it after `install_chunked_code`?

mraszyk: I've removed the new function now: the motivation was to avoid refactoring of the existing (quite large) function to install chunked code.

keplervital (approved):

PR #468, opened 2025-01-06
chore(upgrader): remove deprecated log API and stable storage

mraszyk:
This PR removes the deprecated log API and stable memories in the upgrader, reducing its default stable memory footprint from 40MiB to 24MiB. The deprecated upgrader logs [do not get](https://github.com/dfinity/orbit/blob/7fd45ed6e053fdac2cea05f9807ab45a40244c06/core/upgrader/impl/src/services/logger.rs#L239) any new logs since the last upgrader [release](https://github.com/dfinity/orbit/releases/tag/%40orbit%2Fupgrader-v0.1.0) (end of Nov).


olaszakos (approved):
LGTM

PR #469, opened 2025-01-08
chore(ci): support local replica for local deployment

mraszyk:
This PR adds support for local replica in addition to PocketIC for local deployment. The motivation is that PocketIC started by dfx doesn't properly support resuming a local deployment after `dfx stop`.


olaszakos (approved):
works on my end

PR #470, opened 2025-01-08
feat(wallet): searchable address book; pick from address book at transfer

olaszakos:
What changed:
`AddressBookAutocomplete` now lets you search and pick from the entire address book.
New component: AddressInput that lets you input an address, aided with autocomplete from the address book.

### It now displays the address owner.
image

### It displays the address book entries in the dropdown - should we add the metadata/blockchain to the list items for more info?
image

### Searching now works on both the address and the owner field, fetching the search results from the station
image

### Address owner names are resolved on the policy page,
image

### ...and on the request view
image

### It now has an autocomplete from the address book
image

### But it also accepts new addresses
image

### DIFF
--- a/apps/wallet/src/components/accounts/TransferForm.vue
+++ b/apps/wallet/src/components/accounts/TransferForm.vue
@@ -10,17 +10,27 @@
       :readonly="isViewMode"
       data-test-id="transfer-form-transfer-id"
     />
-    <VTextField
+    <AddressInput
+      v-if="!isViewMode"
       v-model="model.to"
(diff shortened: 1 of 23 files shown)


keplervital (approved):
great work

keplervital:
> It displays the address book entries in the dropdown - should we add the metadata/blockchain to the list items for more info?

i would not add it for now and keep the list simple, the address format should already be a sort of indicator and also the context from where the user is performing the search (e.g. from the transfer dialog of an ICRC-1 asset)

PR #474, opened 2025-01-10
feat(upgrader): new stable memory layout

mraszyk:
This PR introduces a new stable memory layout for the upgrader:
- the bucket size is reduced from the default 8MiB to 1MiB;
- the target canister id is bundled with the disaster recovery state (and a new stable memory version field) into a single stable memory region.

The above two changes reduce the default upgrader stable memory footprint from 3 x 8MiB down to 2 x 1MiB.


olaszakos (approved):
LGTM

PR #481, opened 2025-01-15
feat(wallet): requesters can cancel their requests

olaszakos:
When a request is pending approval the requester can cancel it from the request dialog.

image


jedna (approved):
LGTM

PR #482, opened 2025-01-21
feat(station): configurable station initialization

keplervital:


core/control-panel/impl/src/services/deploy.rs
```diff
@@ -118,11 +118,16 @@ impl DeployService {
         })?;
 
         // The initial admins added to the station.
-        let admins = input
+        let intial_users = input
             .admins
             .iter()
-            .map(|admin| station_api::AdminInitInput {
-                identity: admin.identity,
+            .map(|admin| station_api::UserInitInput {
```
mraszyk, line 124:
```suggestion
            .map(|user| station_api::UserInitInput {
```

core/station/api/spec.did
```diff
@@ -2770,26 +2785,102 @@ type InitAssetInput = record {
...
+  // The initial users to create.
+  users : vec InitUserInput;
+  // The quorum for the initial approval policy.
+  quorum : opt nat16;
+  // The initial entries to create.
+  entries: opt InitalEntries;
 };
 
+
+
```
mraszyk, line 2883:
```suggestion
```

core/station/api/spec.did
```diff
@@ -2770,26 +2785,102 @@ type InitAssetInput = record {
...
+// The init type for initializing the permissions when first creating the canister.
+type InitPermissionInput = record {
+  // The resource that the permission is for.
+  resource : Resource;
+  // The allow rules for who can access the resource.
+  allow : Allow;
+};
+
+// The init type for adding a request approval policies when initializing the canister for the first time.
```
mraszyk, line 2828:
```suggestion
// The init type for adding a request approval policy when initializing the canister for the first time.
```

core/station/api/spec.did
```diff
@@ -2770,26 +2785,102 @@ type InitAssetInput = record {
...
+  // The request specifier that identifies for what operation this policy is for (e.g. "transfer").
+  specifier : RequestSpecifier;
+  // The rule to use for the request approval evaluation (e.g. "quorum").
+  rule : RequestPolicyRule;
+};
+
+// The init type for adding a named rule when initializing the canister for the first time.
+type InitNamedRuleInput = record {
+  // The id of the named rule.
+  id : UUID;
```
mraszyk, line 2841: Why is this UUID not optional (unlike, e.g., for request policy above)?

core/station/api/spec.did
```diff
@@ -2770,26 +2785,102 @@ type InitAssetInput = record {
...
+  name : text;
+  // The identities of the user.
+  identities : vec UserIdentityInput;
+  // The user groups to associate with the user (optional).
+  // If not provided it defaults to the `admin` group if default user groups are created.
+  groups : opt vec UUID;
+  // The status of the user (e.g. `Active`).
+  //
+  // If not provided the default status is `Active` when there is at least
+  // one identity ot `Inactive` otherwise.
```
mraszyk, line 2816:
```suggestion
  // If not provided the default status is `Active` when there is at least
  // one identity or `Inactive` otherwise.
```
Why not always default to `Active`?

core/station/api/src/system.rs
```diff
@@ -59,11 +62,46 @@ pub struct SystemInfoResponse {
 }
 
 #[derive(CandidType, serde::Serialize, Deserialize, Clone, Debug)]
-pub struct AdminInitInput {
+pub struct UserInitInput {
```
mraszyk, line 65: Maybe InitUserInput to be consistent with InitUserGroupInput?

core/station/impl/src/models/request_policy_rule.rs
```diff
@@ -44,6 +44,19 @@ pub enum RequestPolicyRule {
...
 
+impl RequestPolicyRule {
+    pub fn has_named_rule_id(&self, named_rule_id: &NamedRuleId) -> bool {
+        match self {
+            RequestPolicyRule::NamedRule(id) => id == named_rule_id,
+            RequestPolicyRule::And(rules) | RequestPolicyRule::Or(rules) => rules
+                .iter()
+                .any(|rule| rule.has_named_rule_id(named_rule_id)),
+            RequestPolicyRule::Not(rule) => rule.has_named_rule_id(named_rule_id),
+            _ => false,
```
mraszyk, line 55: Could we please expand the wildcard so that the compiler forces us to update this definition upon changes to `RequestPolicyRule`?

tests/integration/src/utils.rs
```diff
@@ -38,6 +38,7 @@ use upgrader_api::{
 use uuid::Uuid;
 
 pub const ADMIN_GROUP_ID: Uuid = Uuid::from_u128(302240678275694148452352); // very first uuidv4 generated
+pub const OPERATOR_GROUP_ID: Uuid = Uuid::from_u128(302240678275694148452353); // very first uuidv4 generated
```
mraszyk, line 41:
```suggestion
pub const OPERATOR_GROUP_ID: Uuid = ADMIN_GROUP_ID; // very first uuidv4 generated
```
if possible

tests/integration/src/install_tests.rs
```diff
@@ -0,0 +1,961 @@
...
+        {
+            return Err(format!("asset {} does not match expected asset", asset.id));
+        }
+    }
+
+    Ok(())
+}
+
+fn compare_arrays<T: PartialEq>(a: &[T], b: &[T]) -> bool {
+    a.len() == b.len() && a.iter().all(|item| b.contains(item))
```
mraszyk, line 853:
```suggestion
    a.len() == b.len() && a.iter().all(|item| b.contains(item)) && b.iter().all(|item| a.contains(item))
```
to cover the case of duplicate entries

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+                }
+                Err(e) => Err(e)?,
+                Ok(_) => {}
+            }
+        }
+
+        Ok(())
+    }
+
     /// Registers the newly added admins of the canister.
```
mraszyk, line 862:
```suggestion
    /// Registers the newly added users of the canister.
```

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
-                identities: vec![admin.identity.to_owned()],
-                groups: vec![ADMIN_GROUP_ID.to_owned()],
-                name: admin.name.to_owned(),
-                status: UserStatus::Active,
-            })?;
+    pub fn set_initial_users(
+        users: Vec<UserInitInput>,
+        default_groups: &[UUID],
+    ) -> Result<(), ApiError> {
+        print(format!("Registering {} admin users", users.len()));
```
mraszyk, line 867:
```suggestion
        print(format!("Registering {} users", users.len()));
```

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+                        .map(UserStatus::from)
+                        .unwrap_or(UserStatus::Active),
+                },
+                user_id,
+            )?;
 
             print(&format!(
-                "Added admin user with principal {} and user id {}",
-                admin.identity.to_text(),
+                "Added admin user with principals {:?} and user id {}",
```
mraszyk, line 903:
```suggestion
                "Added user with principals {:?} and user id {}",
```

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+        });
+
+        for (input, request_policy_id) in add_request_policies {
+            REQUEST_POLICY_SERVICE.add_request_policy_with_id(input, request_policy_id)?;
+        }
+
+        Ok(())
+    }
+
+    // Registers the initial accounts of the canister during the canister initialization.
```
mraszyk, line 814:
```suggestion
    // Registers the initial assets of the canister during the canister initialization.
```

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+                    *HelperMapper::to_uuid(asset.id.clone())
+                        .expect("Invalid UUID")
+                        .as_bytes(),
+                )
+            })
+            .collect::<Vec<(AddAssetOperationInput, UUID)>>();
+
+        for (new_asset, with_asset_id) in add_assets {
+            match ASSET_SERVICE.create(new_asset, Some(with_asset_id)) {
+                Err(ApiError { code, details, .. }) if &code == "ALREADY_EXISTS" => {
```
mraszyk, line 847: How can an asset exist upon canister initialization? Because it is specified multiple times in the input (this sounds like an invalid input)?

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+        // sorting criteria:
+        // - if a policy depends on another policy, the dependent policy should be added first
+        // - keep the original order of the policies otherwise
+        add_request_policies.sort_by(|a, b| {
+            if let Some(a_id) = &a.1 {
+                if specifier_has_reference_to_policy_id(&b.0.specifier, a_id) {
+                    return Ordering::Less;
+                }
+            }
+            Ordering::Greater
```
mraszyk, line 804: I'm not sure this result leads to keeping the original order: shouldn't we compare the original indices here?

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+        specifier: &RequestSpecifier,
+        policy_id: &UUID,
+    ) -> bool {
+        match specifier {
+            RequestSpecifier::EditRequestPolicy(resource_ids)
+            | RequestSpecifier::RemoveRequestPolicy(resource_ids) => match resource_ids {
+                ResourceIds::Any => false,
+                ResourceIds::Ids(ids) => ids.contains(policy_id),
+            },
+            _ => false,
```
mraszyk, line 770: Let's please unfold the wildcard so that the compiler forces us to update this definition upon changes to `RequestSpecifier` type.

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+            .collect::<Result<Vec<_>, _>>()?;
+
+        // sorting criteria:
+        // - if a policy depends on another policy, the dependent policy should be added first
+        // - keep the original order of the policys otherwise
+        add_named_rules.sort_by(|a, b| {
+            if b.0.rule.has_named_rule_id(&a.1) {
+                Ordering::Less
+            } else {
+                Ordering::Greater
```
mraszyk, line 718: I'm not sure this result leads to keeping the original order: shouldn't we compare the original indices here?

core/station/impl/src/services/system.rs
```diff
@@ -291,20 +287,44 @@ impl SystemService {
...
+                                    Some(init_with_permissions.permissions),
+                                )
+                            })
+                            .collect(),
+                        &assets,
+                        quorum,
+                    )
                     .await?;
+                }
+                _ => {}
```
mraszyk, line 327: Let's unfold the wildcard, please.

core/station/impl/src/services/system.rs
```diff
@@ -291,20 +287,44 @@ impl SystemService {
...
-                print("Adding initial assets");
-                install_canister_handlers::set_initial_assets(assets).await?;
-            }
-
-            // if provided, creates the initial accounts
-            if let Some(accounts) = init.accounts {
-                print("Adding initial accounts");
-                install_canister_handlers::set_initial_accounts(accounts, &init.assets, quorum)
+            let initial_user_count = init.users.len() as u16;
+            let quorum = calc_initial_quorum(initial_user_count, init.quorum);
```
mraszyk, line 291: Why is the quorum based on the number of all initial users (incl. non-admins)?

core/station/impl/src/services/system.rs
```diff
@@ -291,20 +287,44 @@ impl SystemService {
...
+
+            match init.entries {
+                Some(InitialEntries::WithDefaultPolicies { accounts, assets }) => {
+                    print("Adding initial accounts");
```
mraszyk, line 295: the debug print is missing in the subsequent match arm

core/station/impl/src/services/system.rs
```diff
@@ -386,24 +406,73 @@ impl SystemService {
...
-            return Err(SystemError::TooManyAdminsSpecified {
+        if input.users.len() > u16::MAX as usize {
+            return Err(SystemError::TooManyUsersSpecified {
                 max: u16::MAX as usize,
             })?;
         }
 
-        // adds the default admin group
-        init_canister_sync_handlers::add_initial_groups();
+        let admin_quorum = calc_initial_quorum(input.users.len() as u16, input.quorum);
```
mraszyk, line 419: looks like user quorum

core/station/impl/src/services/system.rs
```diff
@@ -575,69 +662,257 @@ mod init_canister_sync_handlers {
...
+                    groups: user
+                        .groups
+                        .map(|ids| {
+                            ids.into_iter()
+                                .map(|id| {
+                                    HelperMapper::to_uuid(id.clone())
+                                        .map(|uuid| uuid.as_bytes().to_owned())
+                                })
+                                .collect::<Result<Vec<_>, _>>()
+                                .unwrap_or_else(|_| default_groups.to_vec())
```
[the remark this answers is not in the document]
mraszyk: At other place we panic in such a case.

mraszyk:
What happens if the init args contain the same UUID multiple times? I'm not sure if the code checks for duplicate UUIDs in the input.

tests/integration/src/install_tests.rs
```diff
@@ -0,0 +1,961 @@
...
+            groups: Some(vec![ADMIN_GROUP_ID.hyphenated().to_string()]),
+            id: None,
+            status: Some(station_api::UserStatusDTO::Inactive),
+        },
+        UserInitInput {
+            identities: vec![UserIdentityInput {
+                identity: Principal::from_slice(&[3; 29]),
+            }],
+            name: "other-user".to_string(),
+            groups: Some(vec![]),
```
mraszyk, line 70: What if an invalid group ID was specified here?

tests/integration/src/utils.rs
olaszakos: Here's the comment that is wrong, the operator group id is the 2nd UUID.

core/station/api/spec.did
olaszakos: Not sure what use case we're covering here, it was added by @keplervital . @keplervital can you please elaborate.

core/station/api/spec.did
olaszakos: The idea is that named rules are only useful if they are referenced, and you can only reference one if you're providing a UUID. Do you think it would be better to go for consistency rather than practicality? I don't feel strongly about this.

olaszakos:
> What happens if the init args contain the same UUID multiple times? I'm not sure if the code checks for duplicate UUIDs in the input.

Since the service layer is used for initial entry creation, it's up to the validation implemented there. Thanks for pointing this out, because the newly added `create_xyz_with_id` functions didn't check for duplicate ID's, so the service function would just overwrite those possibly creating invalid state.

core/station/impl/src/services/system.rs
olaszakos: Before this update the default assets (only ICP) were created even when installing with disaster recovery. This can now be removed since the default assets are only added if no initial entries are specified.

core/station/api/spec.did
olaszakos: Changed all id's to optional for consistency

tests/integration/src/install_tests.rs
olaszakos: It will fail with malformed UUID, or if it's valid UUID, it'll fail if it doesn't exist as a group

core/station/impl/src/services/system.rs
olaszakos: replacing `Greater` with `Equal`, since the sort is stable, it should keep the original order for independent named rules/policies

core/station/impl/src/services/system.rs
olaszakos: replacing Greater with Equal, since the sort is stable, it should keep the original order for independent named rules/policies

mraszyk: Good idea!

tests/integration/src/install_tests.rs
mraszyk: Is there a test for this behavior?

core/station/impl/src/services/system.rs
olaszakos: Refactored the quorum input so there is required admin_quorum and operator_quorum when default policies are created.

core/station/impl/src/services/system.rs
olaszakos: I changed `Greater` to `Equal` which should preserve the creation order.

core/station/api/spec.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
...
+  id : opt UUID;
+  // The request specifier that identifies for what operation this policy is for (e.g. "transfer").
+  specifier : RequestSpecifier;
+  // The rule to use for the request approval evaluation (e.g. "quorum").
+  rule : RequestPolicyRule;
+};
+
+// The init type for adding a named rule when initializing the canister for the first time.
+type InitNamedRuleInput = record {
+  // The id of the named rule.
```
mraszyk, line 2840: Should we remark that a new UUID will be generated if none is provided (analogously to, e.g., `InitRequestPolicyInput`)?

core/station/api/spec.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
...
+  resource : Resource;
+  // The allow rules for who can access the resource.
+  allow : Allow;
+};
+
+// The init type for adding a request approval policy when initializing the canister for the first time.
+type InitRequestPolicyInput = record {
+  // The id of the request policy, if not provided a new UUID will be generated.
+  id : opt UUID;
+  // The request specifier that identifies for what operation this policy is for (e.g. "transfer").
```
mraszyk, line 2832:
```suggestion
  // The request specifier that identifies what operation this policy is for (e.g. "transfer").
```

core/station/api/spec.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
...
+// The users to create when initializing the canister for the first time.
+type InitUserInput = record {
+  // The id of the user, if not provided a new UUID will be generated.
+  id : opt UUID;
+  // The name of the user.
+  name : text;
+  // The identities of the user.
+  identities : vec UserIdentityInput;
+  // The user groups to associate with the user (optional).
+  // If not provided it defaults to the `admin` group if default user groups are created,
```
mraszyk, line 2810:
```suggestion
  // If not provided it defaults to the `Admin` group if default user groups are created,
```
as the doc comment of `InitialConfig` defines the admin group as `Admin` (upper-case)

core/station/api/spec.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
...
+  // i.e., when the field `entries` in `SystemInit` is `null` or has the form of `WithDefaultPolicies`.
```
mraszyk, line 2811:
```suggestion
  // i.e., when the field `entries` in `SystemInit` is `null` or has the form of `WithAllDefaults` or `WithDefaultPolicies`.
```

core/station/api/spec.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
   metadata : vec AssetMetadata;
 };
 
-// The init configuration for the canister.
+// The input type for creating a user group when initializing the canister for the first time.
+type InitUserGroupInput = record {
+  // The id of the user group.
```
mraszyk, line 2789:
```suggestion
  // The id of the user group, if not provided a new UUID will be generated.
```

core/station/api/src/system.rs
```diff
@@ -97,22 +150,60 @@ pub struct InitAssetInput {
...
+#[derive(CandidType, serde::Serialize, Deserialize, Clone, Debug)]
+pub enum InitialConfig {
+    WithAllDefaults {
+        /// The initial users to create.
+        users: Vec<UserInitInput>,
+        /// The initial admin quorum in the admin level approval rule.
+        admin_quorum: u16,
+        /// The initial operator quorum in the operator level approval rule.
+        operator_quorum: u16,
+    },
```
mraszyk, line 162: Could we avoid this variant by making the assets and accounts optional in the subsequent variant `WithDefaultPolicies`?

core/station/impl/src/services/user_group.rs
```diff
@@ -92,9 +92,29 @@ impl UserGroupService {
...
+    }
+
+    pub async fn create_with_id(
+        &self,
+        input: AddUserGroupOperationInput,
+        with_user_group_id: Option<UUID>,
+    ) -> ServiceResult<UserGroup> {
+        let user_group_uuid = match with_user_group_id {
+            Some(id) => Uuid::from_bytes(id),
+            None => generate_uuid_v4().await,
```
mraszyk, line 105: Why is this async although the other UUID generations (e.g., for users) are sync (`let user_id = with_id.unwrap_or_else(|| *Uuid::new_v4().as_bytes());`)?

core/station/impl/src/services/user.rs
```diff
@@ -110,13 +110,26 @@ impl UserService {
...
+    pub fn add_user_with_id(
+        &self,
+        input: AddUserOperationInput,
+        with_id: Option<UserId>,
+    ) -> ServiceResult<User> {
         for identity in input.identities.iter() {
             self.assert_identity_has_no_associated_user(identity, None)?;
         }
 
+        let user_id = with_id.unwrap_or_else(|| *Uuid::new_v4().as_bytes());
```
mraszyk, line 128: Shouldn't we also check if this user ID already exists (just like for users groups etc.)?

tests/integration/src/install_tests.rs
```diff
@@ -0,0 +1,1009 @@
...
+            .assets
+            .iter()
+            .find(|asset| asset.name == expected_asset.name)
+            .ok_or(format!("asset {} not found", expected_asset.name))?;
+
+        if expected_asset
+            .id
+            .as_ref()
+            .map(|id| id != &asset.id)
+            .unwrap_or(false)
```
mraszyk, line 883: Do we want to ignore the case of no expected asset ID?

tests/integration/src/install_tests.rs
```diff
@@ -0,0 +1,1009 @@
...
+            .unwrap_or(default_groups)
+            .iter()
+            .all(|group| user.groups.iter().any(|user_group| &user_group.id == group));
+
+        let expected_status = expected_user
+            .status
+            .as_ref()
+            .unwrap_or(&UserStatusDTO::Active);
+
+        if mem::discriminant(&user.status) != mem::discriminant(expected_status) {
```
mraszyk, line 732: we could probably also use `matches!(...)`

tests/integration/src/install_tests.rs
```diff
@@ -0,0 +1,1009 @@
...
+                .iter()
+                .any(|user_identity| user_identity == &identity.identity)
+        });
+
+        expected_user
+            .groups
+            .as_ref()
+            .unwrap_or(default_groups)
+            .iter()
+            .all(|group| user.groups.iter().any(|user_group| &user_group.id == group));
```
mraszyk, line 725: this is presumably ignored because it isn't in an assertion (same above)

tests/integration/src/install_tests.rs
```diff
@@ -0,0 +1,1009 @@
...
+            assets: vec![
+                id_1.clone(), // non-existent asset id
+            ],
+            id: Some(id_1.clone()),
+            seed: Uuid::new_v4().to_bytes_le(),
+        }],
+        assets: vec![],
+    }
+})]
+#[should_panic]
```
mraszyk, line 616: Could we make sure that the panic has the expected cause? (E.g., that the panic is not cause by insufficient cycles)

tests/integration/src/install_tests.rs
```diff
@@ -0,0 +1,1009 @@
...
+            allow: allow_custom_user_group_users.clone(),
+        },
+        InitPermissionInput {
+            resource: ResourceDTO::Permission(PermissionResourceActionDTO::Read),
+            allow: allow_custom_user_group_users.clone(),
+        },
+    ];
+
+    let request_policies = vec![
+        // edit specific request policy, in the wrong order on purpose
```
mraszyk, line 247: what does "wrong order" mean? (this test is supposed to succeed)

core/station/impl/src/services/system.rs
```diff
@@ -422,6 +528,9 @@ impl SystemService {
     /// Updates the canister with the given settings.
     ///
     /// Must only be called within a canister post_upgrade call.
+    ///
+    /// This function is not implemented in the original file or the provided code block.
+    /// It's assumed to exist as it's called in the test case.
```
mraszyk, line 533: I couldn't follow this comment.

core/station/impl/src/services/system.rs
```diff
@@ -662,177 +1013,95 @@ mod install_canister_handlers {
...
+                resource: policy.1.to_owned(),
+            })?;
         }
 
         Ok(())
     }
 
+    #[allow(unused)]
     // Registers the initial accounts of the canister during the canister initialization.
+    // Used
```
mraszyk, line 1038: this looks like an unfinished comment

apps/wallet/src/generated/station/station.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
...
+  // The identities of the user.
+  identities : vec UserIdentityInput;
+  // The user groups to associate with the user (optional).
+  // If not provided it defaults to the `admin` group if default user groups are created,
+  // i.e., when the field `entries` in `SystemInit` is `null` or has the form of `WithDefaultPolicies`.
+  groups : opt vec UUID;
+  // The status of the user (e.g. `Active`).
+  //
+  // If not provided the default status is `Active` when there is at least
+  // one identity or `Inactive` otherwise.
```
mraszyk, line 2816: The status seems to be always `Active` by default (independently of identities). (I guess the comment is out-dated.)

core/station/api/spec.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
...
+type InitialConfig = variant {
+  // Initialize the station with default user groups, named rules, policies, permissions, and assets.
+  // This does not create an initial account. 
+  WithAllDefaults : record {
+    // The initial users to create.
+    users : vec InitUserInput;
+    // The initial admin quorum in the admin level approval rule.
+    admin_quorum : nat16;
+    // The initial operator quorum in the operator level approval rule.
+    operator_quorum : nat16;
```
mraszyk, line 2872: Shouldn't we have a pair of fields `admin_users` and `operator_users` instead of just `users` who become admins and the operator group is empty?

core/station/impl/src/services/system.rs
olaszakos: Must have been added by copilot and didn't notice it. Removing it.

core/station/api/src/system.rs
olaszakos: Closing since this was already changed at the time of the comment

core/station/api/spec.did
olaszakos: Updating

core/station/api/spec.did
olaszakos: Thanks

core/station/api/spec.did
olaszakos: Yes, thanks

core/station/api/spec.did
olaszakos: Then we'd need extra input DTO's for users whose user groups are specified through the field name (ie. admin_users or operator_users), and those could not have other user groups, unless we make it complicated by still allowing the user to specify groups. I don't think that helps the user experience.

apps/wallet/src/generated/station/station.did
olaszakos: No no, the active/inactive logic seems to have never actually existed. Adding it, to the service.

core/station/api/src/system.rs
olaszakos: Ah my bad, I misread your suggestion multiple times.

tests/integration/src/install_tests.rs
olaszakos: The expected asset has an optional ID, and if it's None, this check should evaluate to false, so it won't result in an error immediately. Am I missing something?

core/station/impl/src/services/user_group.rs
olaszakos: Not sure, this `generate_uuid_v4` was probably async for a good reason at some point, but now it seems to be all sync. I'd opt out of extending the scope of the PR to clean this up.

core/station/impl/src/services/user.rs
olaszakos: Good catch

core/station/api/src/system.rs
mraszyk: I see the point, but then I wonder if the groups and status should be optional as their defaults are not very intuitive either.

tests/integration/src/install_tests.rs
mraszyk: > it won't result in an error immediately

My question is if this is desirable: why would the expected asset have its ID equal to None?

tests/integration/src/install_tests.rs
olaszakos: It means the policy is referencing another policy ID, that is specifier later in the list. It is meant to test the dependency-sensitive nature of creating the initial policies.

tests/integration/src/install_tests.rs
olaszakos: Does this have a downside?

core/station/api/spec.did
mraszyk: I see: I (again) forgot the semantics of the current users field...

tests/integration/src/install_tests.rs
olaszakos: The expected asset is of type `InitAssetInput`, whose `id` field is optional.

tests/integration/src/install_tests.rs
mraszyk: I don't think so, I just haven't seen this pattern before.

tests/integration/src/install_tests.rs
olaszakos: yes `test_initial_users_with_bad_groups`

apps/wallet/src/generated/station/station.did
olaszakos: I think it's a good addition as an active user might be required to vote, when they can't authenticate themselves.

mraszyk: The alternative is to reject this as invalid input (the advantage being a simpler semantics of the default value).

core/station/api/src/system.rs
olaszakos: Yes I would make those non-optional as well. The user type was added by @keplervital. @keplervital what do you think of making the groups and status non-optional?

apps/wallet/src/generated/station/station.did
olaszakos: There might be value in being able to create a user which doesn't have an identity yet, eg. when configuring a team station.

tests/integration/src/install_tests.rs
olaszakos: nice catch!

tests/integration/src/install_tests.rs
olaszakos: Adding panic reasons

core/station/impl/src/services/system.rs
olaszakos: Looks resolved to me: https://github.com/dfinity/orbit/blob/kepler/pen-411-configurable-station-init/core/station/impl/src/services/system.rs#L1039

mraszyk: I'm sorry, I missed it when looking through the last two commits.

core/station/api/spec.did
```diff
@@ -2770,24 +2784,136 @@ type InitAssetInput = record {
...
+// The users to create when initializing the canister for the first time.
+type InitUserInput = record {
+  // The id of the user, if not provided a new UUID will be generated.
+  id : opt UUID;
+  // The name of the user.
+  name : text;
+  // The identities of the user.
+  identities : vec UserIdentityInput;
+  // The user groups to associate with the user (optional).
+  // If not provided it defaults to the `Admin` group if default user groups are created,
```
keplervital, line 2810: shouldn't it also have the `Operator` group by default?

core/station/impl/src/services/system.rs
```diff
@@ -386,25 +446,71 @@ impl SystemService {
...
+        match &input.initial_config {
+            InitialConfig::WithAllDefaults {
+                admin_quorum,
+                operator_quorum,
+                users,
+            } => {
+                // adds the default admin group
+                init_canister_sync_handlers::add_default_groups();
+                // registers the admins of the canister
+                init_canister_sync_handlers::set_initial_users(users.clone(), &[*ADMIN_GROUP_ID])?;
```
keplervital, line 458: shouldn't we also add the `OPERATOR_GROUP_ID`? because admins should be able to approve things with the operator policy no?

core/station/impl/src/services/system.rs
```diff
@@ -386,25 +446,71 @@ impl SystemService {
...
+                assets,
+                users,
+                admin_quorum,
+                operator_quorum,
+                ..
+            } => {
+                // adds the default admin group
+                init_canister_sync_handlers::add_default_groups();
+                // registers the admins of the canister
+                init_canister_sync_handlers::set_initial_users(users.clone(), &[*ADMIN_GROUP_ID])?;
```
keplervital, line 477: same here

keplervital, line 477: should we fail here if the users have the `groups` filled up and they are not ADMIN_GROUP_ID or OPERATOR_GROUP_ID? i am wondering this because other then those two, with the `WithDefaultPolicies` and `WithAllDefaults` variants there is no way to provide more user groups

core/station/impl/src/services/system.rs
```diff
@@ -575,71 +701,298 @@ mod init_canister_sync_handlers {
...
+                    .map(|id| HelperMapper::to_uuid(id.clone()).map(|uuid| *uuid.as_bytes()))
+                    .transpose();
+
+                named_rule_id.map(|named_rule_id| (input, named_rule_id))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        // sorting criteria:
+        // - if a policy depends on another policy, the dependent policy should be added first
+        // - keep the original order of the policys otherwise
```
keplervital, line 755:
```suggestion
        // - keep the original order of the policies otherwise
```

core/station/api/src/system.rs
keplervital: let's make the `user_status` non optional then, but for groups i think it's more problematic, because ideally when using those `WithDefaults` and `WithDefaultPolicies` variant we would add the default group_ids for admin and operator, so if we want to avoid any confusion then we might need to have yet another UserInput that that is used for those variants that instead of and array of groups would have a boolean field like `is_admin`

olaszakos (approved):

PR #485, opened 2025-01-22
feat(control-panel): reduce initial station and upgrader cycles to 1T

mraszyk:
This PR reduces the initial station and upgrader cycles to 1T, respectively.

### DIFF
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -2583,7 +2583,10 @@ type SystemUpgraderInput = variant {
   // An existing upgrader canister.
   Id : principal;
   // Creates and deploys a new canister.
-  WasmModule : blob;
+  Deploy : record {
+    wasm_module : blob;
+    initial_cycles : nat;
+  };
 };
 
 // The initial accounts to create when initializing the canister for the first time, e.g., after disaster recovery.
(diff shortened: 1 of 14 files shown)


core/station/api/spec.did
```diff
@@ -2583,7 +2583,10 @@ type SystemUpgraderInput = variant {
   // An existing upgrader canister.
   Id : principal;
   // Creates and deploys a new canister.
-  WasmModule : blob;
+  Deploy : record {
+    wasm_module : blob;
+    initial_cycles : nat;
```
keplervital, line 2588: maybe make it optional and assume a value by default (e.g. 1.5T)?

keplervital (approved):

core/station/api/spec.did
mraszyk: this value must also cover code installation and thus 1.5T might not be enough on the fiduciary subnet; why would you prefer an option? for the sake of disaster recovery?

mraszyk:
@keplervital Also added this safe-guard: https://github.com/dfinity/orbit/pull/485/commits/22862801f454ea57ed86820b03be3f0dee5017b2

core/station/api/spec.did
mraszyk: As discussed in a meeting, I've made the upgrader initial cycles optional with a default value of 2.5T.

keplervital (approved):

PR #486, opened 2025-01-24
feat: docs portal

jedna:


docs/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "npm": "please use pnpm",
+    "yarn": "please use pnpm"
+  },
+  "scripts": {
+    "dev": "astro dev",
+    "start": "astro dev",
+    "build": "astro build",
+    "preview": "astro preview",
+    "astro": "astro",
+    "lint": "concurrently -n prettier,eslint -c auto \"prettier --ignore-path ../.prettierignore --check .\" \"eslint --ext .js,.vue,.ts,.cjs .\"",
```
olaszakos, line 23: does this work with .mdx and .astro?

docs/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "format": "concurrently -n prettier,eslint -c auto \"prettier --ignore-path ../.prettierignore --write .\" \"eslint --ext .js,.vue,.ts,.cjs --fix .\""
+  },
+  "dependencies": {
```
olaszakos, line 26: pin dependencies?

canister_ids.json
```diff
@@ -9,6 +9,9 @@
     "production": "5mxvd-qaaaa-aaaal-ajbkq-cai",
     "testing": "lotbt-qqaaa-aaaal-aduzq-cai"
   },
+  "docs_portal": {
+    "production": "bp6mw-eqaaa-aaaac-ahroq-cai"
```
olaszakos, line 13: who controls this?

olaszakos:
Also consider adding cache control headers so that the fonts dont flash on every page change.

docs/package.json
jedna: yes, added in https://github.com/dfinity/orbit/pull/486/commits/5f158885ccb54e5275b12664b68be3f362719f3e

canister_ids.json
jedna: currently I do, but will extend the ownership to @keplervital and whoever needs it until we have an automated deployment in place

jedna:
Note: astro eslint plugin bumped the version of eslint, causing warnings due to deprecated configuration in our repo. It is working with the `ESLINT_USE_FLAT_CONFIG=false` flag, but the configuration must be updated with eslint v10.

olaszakos (approved):

docs/src/content/docs/developers/getting-started.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Getting Started
```
keplervital, line 2: i would maybe structure this differently, wdyt of something like this?

```
> Overview
> User
> Integrator
> Contributor
```

docs/src/content/nav/en.ts
```diff
@@ -0,0 +1,4 @@
+export default {
+  users: 'User documentation',
+  developers: 'Developer documentation',
```
keplervital, line 3: this depends on the decision above on the sections

keplervital (approved):

docs/src/content/docs/developers/getting-started.md
jedna: let's think of the groups once we better understand the content. I suggest we first focus on "User" for the public release and add more afterward. WDYT?

docs/README.md
```diff
@@ -0,0 +1,54 @@
...
+| `npm install`             | Installs dependencies                            |
+| `npm run dev`             | Starts local dev server at `localhost:4321`      |
+| `npm run build`           | Build your production site to `./dist/`          |
+| `npm run preview`         | Preview your build locally, before deploying     |
+| `npm run astro ...`       | Run CLI commands like `astro add`, `astro check` |
+| `npm run astro -- --help` | Get help using the Astro CLI                     |
+
+## 👀 Want to learn more?
+
+Check out [Starlight’s docs](https://starlight.astro.build/), read [the Astro documentation](https://docs.astro.build), or jump into the [Astro Discord server](https://astro.build/chat).
```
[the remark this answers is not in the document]
jedna: agree, we're Seasoned Astronauts

PR #487, opened 2025-01-27
feat(control-panel): user can deploy 2 stations per day

mraszyk:
This PR updates the rate limiting policy in the control panel to allow every user to deploy 2 stations per day (instead of 3 stations in total).


keplervital (approved):
nice 🚀

PR #488, opened 2025-01-27
feat(wallet): named rules page + add/remove/edit rules

olaszakos:


apps/wallet/src/components/users/InteractiveNamedRule.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <template v-if="loading">
+    <VProgressCircular indeterminate />
+  </template>
+  <template v-else-if="error">
+    {{ id }}
```
keplervital, line 6: should we give any indication of the error as well?

apps/wallet/src/components/request-policies/rule/NamedRule.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <div class="d-flex align-center justify-start">
+    {{ $t('request_policies.rule.named_rule') }}
+    <template v-if="loading">
+      <VProgressCircular indeterminate />
+    </template>
+    <template v-else-if="error">
+      {{ $t('request_policies.rule.named_rule') }}
```
keplervital, line 8: the error should be surfaced here instead of repeating the same thing right? or the namedRuleId

apps/wallet/src/composables/request-policies.composable.ts
```diff
@@ -45,3 +102,368 @@ export const useUserSpecifierSelectorItems = (): ComputedRef<SelectItem[]> => {
...
+      rule: PopulatedRule;
+    }
+  | {
+      kind: 'NamedRule';
+      name?: string;
+      id: UUID;
+      rule?: PopulatedRule;
+    };
+
+async function populateUserSpecifier(
```
keplervital, line 166: we should improve this in a separate PR to ensure we keep some caching on the get operations to avoid too many network requests

keplervital (approved):

apps/wallet/src/composables/request-policies.composable.ts
olaszakos: Agreed, a cache in the app store with a short TTL would cover it I think.

keplervital (approved):

PR #489, opened 2025-01-29
feat(control-panel): rate limiting to at most 100 stations per day

mraszyk:
This PR adds rate limiting to at most 100 stations deployed jointly by all users per day.

### DIFF
--- a/core/control-panel/impl/src/controllers/station.rs
+++ b/core/control-panel/impl/src/controllers/station.rs
@@ -145,7 +145,6 @@ impl StationController {
         let ctx = CallContext::get();
         self.user_service
             .can_deploy_station(&ctx)
-            .await
             .map(|can_deploy_station| can_deploy_station.into())
     }
 }
--- a/core/control-panel/impl/src/controllers/station.rs
+++ b/core/control-panel/impl/src/controllers/station.rs
@@ -1,8 +1,8 @@
 //! Station services.
+use crate::core::canister_config;
 use crate::core::middlewares::use_canister_call_metric;
-use crate::errors::UserError;
+use crate::errors::{DeployError, UserError};
 use crate::mappers::user_station::UpdateUserStationInputInto;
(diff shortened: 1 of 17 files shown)


core/control-panel/impl/src/controllers/station.rs
```diff
@@ -130,7 +131,16 @@ impl StationController {
     async fn deploy_station(&self, input: DeployStationInput) -> ApiResult<DeployStationResponse> {
         let ctx = CallContext::get();
         let _lock = STATE
-            .with(|state| CallerGuard::new(state.clone(), ctx.caller(), None))
+            .with(|state| {
+                CallerGuard::new(
+                    state.clone(),
+                    ctx.caller(),
+                    CallerGuardParams {
+                        max_concurrency: Some(MAX_DEPLOYED_STATIONS_PER_DAY),
```
mraszyk, line 139: We should decrease MAX_DEPLOYED_STATIONS_PER_DAY by the current number of already deployed stations on the same day.

core/station/impl/src/services/account.rs
```diff
@@ -553,7 +553,10 @@ impl AccountService {
                                 CallerGuard::new(
                                     state.clone(),
                                     balance_update_guard_key,
-                                    Some(time() + Duration::from_secs(5 * 60).as_nanos() as u64),
+                                    CallerGuardParams {
+                                      max_concurrency: None,
+                                      expires_at_ns: Some(time() + Duration::from_secs(5 * 60).as_nanos() as u64)
+                                    },
```
keplervital, line 559: nit: it would be nice to have this with a builder pattern and derive default, this way you would be able to write this as

```suggestion
CallerGuardParams::default()
  .with_expires_at_ns(time() + Duration::from_secs(5 * 60).as_nanos() as u64),
```

keplervital (approved):

PR #490, opened 2025-01-30
feat(control-panel): remove waiting list

mraszyk:
This PR removes the waiting list and sets the stable memory version in `CanisterConfig::default()` so that memory migration is not triggered for upgrades from new control panel deployments.

### DIFF
--- a/core/control-panel/impl/src/core/config.rs
+++ b/core/control-panel/impl/src/core/config.rs
@@ -37,7 +37,7 @@ impl Default for CanisterConfig {
             station_wasm_module: vec![],
             station_wasm_module_extra_chunks: None,
             last_upgrade_timestamp: time(),
-            version: None,
+            version: Some(SYSTEM_VERSION.to_string()),
             global_rate_limiter: RateLimiter::new_global(),
         }
     }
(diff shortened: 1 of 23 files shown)


core/control-panel/impl/src/core/config.rs
```diff
@@ -37,7 +37,7 @@ impl Default for CanisterConfig {
             station_wasm_module: vec![],
             station_wasm_module_extra_chunks: None,
             last_upgrade_timestamp: time(),
-            version: None,
+            version: Some(SYSTEM_VERSION.to_string()),
```
olaszakos, line 40: was this a bug that this was missing?

olaszakos (approved):

core/control-panel/impl/src/core/config.rs
mraszyk: I believe this is a bug that doesn't matter in production as we don't deploy new control panels from scratch.

keplervital: also in prod we always use the `CanisterConfig::new(args)` instead of `CanisterConfig::default()`, and new does include the version.

PR #492, opened 2025-02-03
fix(station): docker build

olaszakos:
Fixes 2 issues:

1. pnpm doesnt work - https://github.com/nodejs/corepack/issues/612 -> solved by installing latest (pinned) corepack manually
2. deprecated `upload-artifact@v3` -> followed the [migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) to update to @v4

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -78,9 +78,9 @@ jobs:
           mkdir -p artifacts
           mv tests/integration/wasms/* artifacts
       - name: 'Upload Artifacts'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: integration-tests
+          name: integration-tests-${{ matrix.canister }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -108,12 +108,21 @@ jobs:
...
+    needs: [build-wasms, download-canisters]
+    steps:
+      - name: Merge Artifacts
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: integration-tests
+          pattern: integration-tests-*
   integration-tests:
     name: 'integration-tests:required'
     needs: [build-wasms, download-canisters]
```
keplervital, line 128: wouldn't this need to depend on `merge-artifacts` now?

keplervital (approved):

.github/workflows/tests.yaml
olaszakos: indeed, nice catch!

PR #494, opened 2025-02-04
chore(wallet): verify locale keys at dev/build

olaszakos:
Added a new build plugin that checks the secondary locales if they contain all keys the English locale does. Outputs a warning in dev, fails the build in prod. Also fixed the missing keys.


keplervital (approved):
LGTM

PR #511, opened 2025-02-20
chore(station): add POCKET_IC_BIN to build.rs

olaszakos:


tests/integration/build.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    let env_vars = vec![(
+        "POCKET_IC_BIN",
+        format!("{}/pocket-ic", std::env::current_dir().unwrap().display()),
```
mraszyk, line 4: is `std::env::current_dir()` the directory from where the user runs cargo or the directory containing the Cargo.toml file for this Rust source?

olaszakos: According to the docs, the current directory for the build script is always the root directory of the package it belongs to: https://doc.rust-lang.org/cargo/reference/build-scripts.html#inputs-to-the-build-script, which is where the pocket-ic binary gets downloaded to.

mraszyk: Thank you for confirming!

keplervital (approved):

PR #526, opened 2025-03-03
feat(upgrader): support restoring snapshots in disaster recovery via upgrader

mraszyk:


core/upgrader/impl/src/services/disaster_recovery.rs
```diff
@@ -306,6 +308,22 @@ impl DisasterRecoveryService {
...
+            }
+            StationRecoveryRequestOperation::Restore(snapshot) => {
+                installer.stop(station_canister_id).await?;
+
+                let snapshot_args = LoadCanisterSnapshotArgs {
+                    canister_id: station_canister_id,
+                    snapshot_id: snapshot.snapshot_id,
+                    sender_canister_version: Some(canister_version()),
+                };
+                load_canister_snapshot(snapshot_args)
```
olaszakos, line 321: Please test this path in unit tests. Since caling directly is not testable, consider adding this method to the `InstallCanister` trait, and maybe rename it to `ManageCanister`, and call it through that trait.

mraszyk: I have integration tests for this code path. Are they not enough?

olaszakos: Well in this case there is indeed not a lot to test, it's mainly to get full coverage and to continue the existing pattern. Deferrig to your judgement.

olaszakos (approved):

core/upgrader/impl/src/services/disaster_recovery.rs
mraszyk: My reasoning was that the existing pattern allows unit testing general disaster recovery properties and the new integration tests would provide coverage of the new code path specifically.

PR #534, opened 2025-03-05
feat(upgrader): prune resources in disaster recovery via upgrader

mraszyk:


core/upgrader/impl/src/mappers/disaster_recovery.rs
```diff
@@ -134,6 +181,22 @@ impl From<&StationRecoveryRequestOperation> for upgrader_api::StationRecoveryReq
                     },
                 )
             }
+            StationRecoveryRequestOperation::Prune(ref prune) => {
```
keplervital, line 184: these mappers would benefit from some unit tests

mraszyk: What would you assert in those unit tests? Just the result of the conversion?

mraszyk: This is what the test would like for checking conversion of pruning operations (and only from/to upgrader_api types):
```
#[test]
    fn api_types() {
        assert!(matches!(
            (&StationRecoveryRequestOperation::Prune(
                StationRecoveryRequestPruneOperation::Snapshot(vec![0x42],)
            ))
                .into(),
            upgrader_api::StationRecoveryRequestOperation::Prune(
                upgrader_api::StationRecoveryRequestPruneOperation::Snapshot(_)
            ),
        ));
        assert!(matches!(
            (&StationRecoveryRequestOperation::Prune(
                StationRecoveryRequestPruneOperation::ChunkStore
            ))
                .into(),
            upgrader_api::StationRecoveryRequestOperation::Prune(
                upgrader_api::StationRecoveryRequestPruneOperation::ChunkStore
            ),
        ));
        assert!(matches!(
            (&StationRecoveryRequestOperation::Prune(StationRecoveryRequestPruneOperation::State))
                .into(),
            upgrader_api::StationRecoveryRequestOperation::Prune(
                upgrader_api::StationRecoveryRequestPruneOperation::State
            ),
        ));

        assert!(matches!(
            upgrader_api::RequestDisasterRecoveryInput::Prune(
                upgrader_api::RequestDisasterRecoveryPruneInput::Snapshot("42".to_string())
            )
            .into(),
            StationRecoveryRequestOperation::Prune(StationRecoveryRequestPruneOperation::Snapshot(
                _
            ))
        ));
        assert!(matches!(
            upgrader_api::RequestDisasterRecoveryInput::Prune(
                upgrader_api::RequestDisasterRecoveryPruneInput::ChunkStore
            )
            .into(),
            StationRecoveryRequestOperation::Prune(
                StationRecoveryRequestPruneOperation::ChunkStore
            )
        ));
        assert!(matches!(
            upgrader_api::RequestDisasterRecoveryInput::Prune(
                upgrader_api::RequestDisasterRecoveryPruneInput::State
            )
            .into(),
            StationRecoveryRequestOperation::Prune(StationRecoveryRequestPruneOperation::State)
        ));
    }
```
The test is very repetitive so I'm not sure about its added value to be honest.

keplervital: Got it, i leave it up to your discretion, I see that we already have integration tests covering this. The main idea here was to add lightweight unit tests to ensure the mappings are correct since we do them manually. 

In Java, we have tools like [MapStruct](https://github.com/mapstruct/mapstruct) that auto-generates these mappings between API types and internal types, which would be ideal if we had something similar in Rust, that way, we could avoid manually adding both the mappings and the unit tests for them.

keplervital (approved):

PR #544, opened 2025-03-14
feat(station): make canister_status endpoint canfund compatible

mraszyk:
This PR makes the `canister_status` endpoint of the station compatible with canfund by aligning its return value with the return value of the management canister.

### DIFF
--- a/apps/wallet/src/generated/station/station.did
+++ b/apps/wallet/src/generated/station/station.did
@@ -3198,11 +3198,6 @@ type CanisterSnapshotsResult = variant {
   Err : Error;
 };
 
-type CanisterStatusResult = variant {
-  Ok : CanisterStatusResponse;
-  Err : Error;
-};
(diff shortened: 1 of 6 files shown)


jedna:
@mraszyk you will need to also update the frontend code to properly read the new response, e.g. https://github.com/dfinity/orbit/blob/b30f1954e9207f2a1fd381b7cec5253596fa3bf8/apps/wallet/src/services/station.service.ts#L769-L779

mraszyk:
> @mraszyk you will need to also update the frontend code to properly read the new response in
> 
> https://github.com/dfinity/orbit/blob/b30f1954e9207f2a1fd381b7cec5253596fa3bf8/apps/wallet/src/services/station.service.ts#L769-L779

Could you please update the frontend code, @jedna or @keplervital ? I'm not familiar with that.

PR #548, opened 2025-04-04
feat(station,upgrader): backup snapshot in system upgrade requests

mraszyk:
This PR adds support for taking a backup snapshot during system upgrade request execution (before performing the actual upgrade). A rolling window of backup snapshots is introduced (currently of length 1 - to be generalized in a follow up PR).


core/station/api/spec.did
```diff
@@ -601,6 +601,10 @@ type SystemUpgradeOperationInput = record {
   module_extra_chunks : opt WasmModuleExtraChunks;
   // The initial argument passed to the new wasm module.
   arg : opt blob;
+  // Determines whether a backup snapshot should be taken (before the upgrade).
+  // If so, the last backup snapshot (if one exists) is atomically replaced
```
olaszakos, line 605: What does "atomically" mean here?

mraszyk: Atomically means that the old snapshot is only deleted if a new one is successfully taken, i.e., you won't end up with no snapshot if taking a new snapshot fails.

libs/orbit-essentials/src/backup_snapshots.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl BackupSnapshots {
+    pub fn new(max_backup_snapshots: usize) -> Self {
+        Self {
+            max_backup_snapshots,
+            backup_snapshots: VecDeque::new(),
+        }
+    }
+
+    pub fn replace_snapshot(&self) -> Option<Vec<u8>> {
```
olaszakos, line 25: I think since "replace" is a verb, it suggests this function will replace a snapshot. I think a more suitable name would be "get_snapshot_to_replace". Same as regarding `SystemInfo::replace_upgrader_backup_snapshot`.

PR #552, opened 2025-04-07
chore(station): perform request validation before execution

mraszyk:
This PR repeats request validation before execution to cover cases when the request becomes invalid while accepting votes.


mraszyk:
TODO:
- add test for executing scheduled requests after their expiry
- refactor expiry check to not include requests that are already scheduled for execution

olaszakos (approved):

PR #553, opened 2025-04-07
fix(station): restrict access to system canister info

mraszyk:
This PR restricts access to reading canister snapshots/status for system canisters by using the permission `Resource::System(SystemResourceAction::SystemInfo)` instead of external canister permissions which are not appropriate for system canisters.

### DIFF
--- a/core/station/impl/src/mappers/authorization.rs
+++ b/core/station/impl/src/mappers/authorization.rs
@@ -385,7 +385,7 @@ impl From<&station_api::CreateRequestInput> for Resource {
 impl From<&CanisterSnapshotsInput> for Resource {
     fn from(input: &CanisterSnapshotsInput) -> Self {
         let canister_id = input.canister_id;
-        if let Ok(()) = EnsureExternalCanister::is_external_canister(canister_id) {
+        if EnsureExternalCanister::is_external_canister(canister_id) {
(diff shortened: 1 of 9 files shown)


core/station/impl/src/mappers/authorization.rs
```diff
@@ -380,6 +382,32 @@ impl From<&station_api::CreateRequestInput> for Resource {
     }
 }
 
+impl From<&CanisterSnapshotsInput> for Resource {
+    fn from(input: &CanisterSnapshotsInput) -> Self {
+        let canister_id = input.canister_id;
+        if let Ok(()) = EnsureExternalCanister::is_external_canister(canister_id) {
```
olaszakos, line 388: using .is_ok() instead would be cleaner IMHO

olaszakos (approved):

core/station/impl/src/mappers/authorization.rs
mraszyk: I went for the latter in this [commit](https://github.com/dfinity/orbit/pull/553/commits/3b00ebe020f57e519cb2ecde600e3114f25f6569) since the main motivation for `let Ok(())` was to prevent accidental changes to the API, e.g., making it return `Result<bool, ...>` and forgetting to consider the inner Boolean value at this call site.

PR #560, opened 2025-04-11
chore(station): bump stable memory to V3

mraszyk:
This PR bumps station's stable memory to V3. The stable memory V2 was generated at the commit 6dce3f7 at which this stable memory V2 was introduced.


mraszyk:
Thanks, @olaszakos , for fixing the migration tests!

PR #563, opened 2025-04-24
chore(ci): bump PocketIC server to v9.0.0 and library to v8.0.0

mraszyk:


tests/integration/src/external_canister_tests.rs
```diff
@@ -1390,29 +1390,6 @@ fn snapshot_external_canister_test() {
     .unwrap();
     assert_eq!(ctr, 4_u32.to_le_bytes());
 
-    // taking another snapshot without specifying a snapshot to replace should fail
```
mraszyk, line 1393: Note for reviewers: this is no longer the case now that the protocol allows up to 10 snapshots, but dropping this edge case should be fine since we have a test for replace_snapshot later.

mraszyk:
Merged as part of https://github.com/dfinity/orbit/pull/565

PR #570, opened 2025-05-12
feat(station): validate AllowListed and AllowListedByMetadata rules

olaszakos:
This PR adds validations to make sure rules containing AllowListed and AllowListedByMetadata can only be applied to Transfer requests by:
- adding specifier-rule compatibility check to RequestPolicy validation
- adding validation to NamedRules to check if any policy directly or indirectly referencing the named rule is compatible with the rule


core/station/impl/src/models/request_policy_rule.rs
```diff
@@ -44,6 +45,22 @@ pub enum RequestPolicyRule {
...
+        match self {
+            RequestPolicyRule::AutoApproved => write!(f, "AutoApproved"),
+            RequestPolicyRule::QuorumPercentage(_, _) => write!(f, "QuorumPercentage"),
+            RequestPolicyRule::Quorum(_, _) => write!(f, "Quorum"),
+            RequestPolicyRule::AllowListedByMetadata(_) => write!(f, "AllowListedByMetadata"),
+            RequestPolicyRule::AllowListed => write!(f, "AllowListed"),
+            RequestPolicyRule::Or(_) => write!(f, "Or"),
+            RequestPolicyRule::And(_) => write!(f, "And"),
+            RequestPolicyRule::Not(_) => write!(f, "Not"),
+            RequestPolicyRule::NamedRule(_) => write!(f, "NamedRule"),
```
mraszyk, line 59: Do we really want to hide the nested rules here?

core/station/impl/src/models/request_policy_rule.rs
```diff
@@ -59,6 +76,19 @@ impl RequestPolicyRule {
...
+    //         return true;
+    //     } else {
+
+    //     match self {
+    //         RequestPolicyRule::NamedRule(id) => {
+
+    //         }
+    //         _ => false
+    //     }
+    // }
```
mraszyk, line 91:
```suggestion
```

core/station/impl/src/models/request_policy.rs
```diff
@@ -53,10 +53,29 @@ impl From<MatchError> for EvaluateError {
     }
 }
 
+// metadata related rules can only be applied to accounts
+pub fn validate_rule_for_specifier(
+    rule: &RequestPolicyRule,
+    specifier: &RequestSpecifier,
+) -> ModelValidatorResult<RequestPolicyError> {
+    match (rule, specifier) {
```
mraszyk, line 61: How about separating the two conditions:
```
let is_metadata_rule = match rule {
  RequestPolicyRule::AllowListed | RequestPolicyRule::AllowListedByMetadata(_) => true,
  _ => false,
};
let is_transfer = match specifier {
  RequestSpecifier::Transfer(_) => true,
  _ => false,
};
if is_metadata_rule && !is_transfer {
  Err(RequestPolicyError::InvalidRuleForSpecifier {
    rule: rule.to_string(),
    specifier: specifier.to_string(),
  })
} else {
  Ok(())
}
```

core/station/impl/src/models/request_policy.rs
```diff
@@ -53,10 +53,29 @@ impl From<MatchError> for EvaluateError {
     }
 }
 
+// metadata related rules can only be applied to accounts
```
mraszyk, line 56: A generic remark: should we avoid wildcards to ensure we for sure catch all rules and specifiers (also in the future)?

core/station/impl/src/models/request_policy.rs
```diff
@@ -53,10 +53,29 @@ impl From<MatchError> for EvaluateError {
...
+    match (rule, specifier) {
+        (RequestPolicyRule::AllowListed, RequestSpecifier::Transfer(_))
+        | (RequestPolicyRule::AllowListedByMetadata(_), RequestSpecifier::Transfer(_)) => Ok(()),
+        (RequestPolicyRule::AllowListed, _) | (RequestPolicyRule::AllowListedByMetadata(_), _) => {
+            Err(RequestPolicyError::InvalidRuleForSpecifier {
+                rule: rule.to_string(),
+                specifier: specifier.to_string(),
+            })
+        }
+        (_, _) => Ok(()),
```
mraszyk, line 70: The cases `AnyOf : vec RequestPolicyRule`, `AllOf : vec RequestPolicyRule`, and `Not : RequestPolicyRule` could contain nested metadata rules that we missed here.

mraszyk: Hence, we need to do a full traversal of `RequestPolicyRule`.

core/station/impl/src/models/named_rule.rs
```diff
@@ -168,6 +197,31 @@ fn validate_description(description: &Option<String>) -> ModelValidatorResult<Na
...
+        for referencing_named_rule in referencing_named_rules.iter() {
+            if policy.rule.has_named_rule_id(referencing_named_rule) {
+                validate_rule_for_specifier(rule, &policy.specifier).map_err(|e| {
+                    NamedRuleError::IncompatibleWithLinkedPolicy {
+                        policy_id: Uuid::from_bytes(policy.id).hyphenated().to_string(),
+                        error: e.to_string(),
+                    }
+                })?;
+            }
+        }
```
mraszyk, line 219:
```suggestion
                validate_rule_for_specifier_with_substitution(policy.rule, &policy.specifier, id, rule).map_err(|e| {
                    NamedRuleError::IncompatibleWithLinkedPolicy {
                        policy_id: Uuid::from_bytes(policy.id).hyphenated().to_string(),
                        error: e.to_string(),
                    }
                })?;
```
where the function `validate_rule_for_specifier_with_substitution` performs a traversal of `RequestPolicyRule` just like `validate_rule_for_specifier`, but whenever it encounters the named rule with `id`, then it continues the traversal on `rule` instead of the current child. This terminates assuming there are no (indirect) circular references (which we should ensure separately).

I pushed a [counterexample](https://github.com/dfinity/orbit/pull/570/commits/430bd790cb6e040950aaa48b7a74ee800d0ab839) why the current algorithm is not correct.

core/station/impl/src/models/request_policy.rs
olaszakos: good catch!

core/station/impl/src/models/request_policy_rule.rs
mraszyk: Maybe just the structure without details in the leaves?

core/station/impl/src/models/request_policy.rs
mraszyk: I see. However, the risk with matching pairs is that we can easily miss some combinations in the product.

core/station/impl/src/models/request_policy_rule.rs
olaszakos: Sounds good

core/station/impl/src/models/request_policy.rs
olaszakos: Addressed in [f3e14ab](https://github.com/dfinity/orbit/pull/570/commits/f3e14ab3f9430db8007f4bfd2a4f4df36703f3a6)

core/station/impl/src/models/request_policy.rs
olaszakos: I removed the tuple match and instead we're just matching for the rule exhaustively. Then the Transfer-only condition is simpler. What do you think? We could still list all request operations to not use the wildcard, maybe it's worth it?

core/station/impl/src/models/request_policy_rule.rs
olaszakos: Removed the dead code.

core/station/impl/src/models/request_policy_rule.rs
olaszakos: Addressed in [e766793](https://github.com/dfinity/orbit/pull/570/commits/e7667933f66cad1d4d6ec6f1b61d54f2f204cf32), I introduced a new field in the error so that it can hold both the offending rule and the full rule.

core/station/impl/src/models/request_policy_rule.rs
```diff
@@ -44,6 +45,77 @@ pub enum RequestPolicyRule {
...
+                write!(f, "Not(")?;
+                rule.fmt_with_context(f, visited)?;
+                write!(f, ")")
+            }
+            RequestPolicyRule::NamedRule(id) => {
+                if !visited.insert(*id) {
+                    return write!(
+                        f,
+                        "NamedRule(CIRCULAR_REFERENCE {})",
+                        Uuid::from_bytes(*id).hyphenated()
```
mraszyk, line 98: I'm not sure if the id is useful here since we don't print the matching id in the parent.

core/station/impl/src/models/request_policy.rs
```diff
@@ -150,4 +222,95 @@ pub mod request_policy_test_utils {
...
+                name: "test_2".to_string(),
+                description: None,
+                rule: RequestPolicyRule::AllowListedByMetadata(MetadataItem {
+                    key: "test".to_string(),
+                    value: "test".to_string(),
+                }),
+            })
+            .expect("Failed to create named rule");
+
+        let tests = [
```
mraszyk, line 250: How about only listing the rules and errors here and later do a "product" test:
```
for (specifier, expect_err) in [(RequestSpecifier::AddAccount, true), (RequestSpecifier::Transfer, false)] {
  for (rule, err) in tests {
    // ...
    let result = policy.validate();
    if expect_err {
      assert_eq!(result, expected);
    } else {
      result.expect("...");
    }
  }
}
```

core/station/impl/src/models/request_policy.rs
mraszyk: The current version looks good; I'd still remove the wildcard though.

core/station/impl/src/models/named_rule.rs
mraszyk: I'd still simplify:
```
/// Validates that the named rule is compatible with the policies that reference it.
/// It traverses all linked policy rules recursively and assumes no circular references.
fn validate_policy_compatibility(
    id: &NamedRuleId,
    rule: &RequestPolicyRule,
) -> ModelValidatorResult<NamedRuleError> {
    let policies = REQUEST_POLICY_REPOSITORY.list();
    for policy in policies {
                validate_rule_for_specifier(&policy.rule, &policy.specifier, &[(id, rule)])
                    .map_err(|e| NamedRuleError::IncompatibleWithLinkedPolicy {
                        policy_id: Uuid::from_bytes(policy.id).hyphenated().to_string(),
                        error: e.to_string(),
                    })?;
    }

    Ok(())
}
```
This way, we only traverse all policy rules once (better worst-case runtime) and the algorithm is simpler (more robust, less changes for logical bugs).

core/station/impl/src/errors/request_policy.rs
```diff
@@ -15,6 +15,13 @@ pub enum RequestPolicyError {
     /// Request policy with id `{id}` already exists.
     #[error(r#"Request policy with id `{id}` already exists."#)]
     IdAlreadyExists { id: String },
+    /// The rule `{invalid_rule}` is invalid for the specifier `{specifier}`.
+    #[error(r#"The rule `{invalid_rule}` is invalid for the specifier `{specifier}`."#)]
+    InvalidRuleForSpecifier {
+        invalid_rule: String,
+        specifier: String,
+        rule: String,
```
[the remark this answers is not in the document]
mraszyk: Or maybe even remove the "root" rule? It can be fetched separately if somebody wants to check it out.

core/station/impl/src/models/request_policy_rule.rs
mraszyk: My suggestion: only print `"NamedRule(CIRCULAR_REFERENCE)"`.

core/station/impl/src/errors/request_policy.rs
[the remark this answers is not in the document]
olaszakos: Changed as discussed.

core/station/impl/src/models/named_rule.rs
olaszakos: Changed as discussed.

core/station/impl/src/models/request_policy.rs
olaszakos: I have removed the wildcard.

core/station/impl/src/models/request_policy.rs
mraszyk: I wanted to increase coverage (without a lot of copying), but I see your point and don't have a strong opinion here.

core/station/impl/src/models/named_rule.rs
olaszakos: I dont think that would be good UX. The user would need to understand the nuances of the policy/named_rule/rule system in order to get past it. Let's discuss with Kepler next week?

mraszyk: > I dont think that would be good UX. The user would need to understand the nuances of the policy/named_rule/rule system in order to get past it.

I agree, but now we risk having invalid stuff in the upgraded station and potentially getting "stuck" if the validation is on a critical path.

> Let's discuss with Kepler next week?

Indeed.

olaszakos: To get stuck what would need to happen? You can always edit a policy rule directly, which should not involve validation of something else. That way you should be able to fix all invalid policies. Then you can fix the named rules if any of them are invalid, since no policies will use them. Can you think of an unrecoverable scenario?

mraszyk: E.g., what if there's no valid policy according to the new rules: will the user still be able to execute a request? I guess there are two options:
- fail in post-upgrade if there's any violation of the new rules;
- add end-to-end tests that a violation of the new rules can be recovered after an upgrade.

keplervital (approved):

PR #576, opened 2025-05-14
feat(station): update Request::last_modification_timestamp on adding approval

olaszakos:
This PR implements the update of Request::last_modification_timestamp on adding an approval.

### DIFF
--- a/core/station/impl/src/services/request.rs
+++ b/core/station/impl/src/services/request.rs
@@ -464,6 +464,8 @@ impl RequestService {
         // Must happen after the approval is added to the request to ensure the approval is counted.
         let maybe_evaluation = request.reevaluate().await?;
 
+        request.last_modification_timestamp = next_time();
+
         self.request_repository
             .insert(request.to_key(), request.to_owned());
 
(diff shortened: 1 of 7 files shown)


core/station/impl/src/services/request.rs
```diff
@@ -464,6 +464,8 @@ impl RequestService {
         // Must happen after the approval is added to the request to ensure the approval is counted.
         let maybe_evaluation = request.reevaluate().await?;
 
+        request.last_modification_timestamp = next_time();
+
         self.request_repository
             .insert(request.to_key(), request.to_owned());
```
mraszyk, line 470: Maybe we could hide these two steps behind a helper and use that helper consistently?

olaszakos: How about something like 

```
impl RequestRepository {
    
    ...
    
    pub fn save_modified(&self, mut request: Request, modified_at: u64) {
        request.last_modification_timestamp = modified_at;
        self.insert(request.to_key(), request.to_owned());
    }
```

olaszakos: In a lot of instances of writing last_modification_timestamp the value is some variable, so yes I think we need to pass in the time.

olaszakos: Addressed the above, also had to tweak the CI for the deterministic build to pass: we got the same issue as here https://github.com/docker-practice/actions-setup-docker/issues/41#issuecomment-2621912932, and the recommended solution to move to `docker/setup-docker-action@v4` fixed it.

mraszyk (approved):

PR #580, opened 2025-05-21
chore(ci): bump uuid crate

mraszyk:
This PR bumps the `uuid` and `getrandom` crates. Because of
```
#[no_mangle]
unsafe extern "Rust" fn __getrandom_v03_custom(
```
multiple versions of the orbit essentials library in the build tree might lead to linking errors.

### DIFF
--- a/libs/orbit-essentials/src/utils/random.rs
+++ b/libs/orbit-essentials/src/utils/random.rs
@@ -27,8 +27,6 @@ unsafe extern "Rust" fn __getrandom_v03_custom(
 ) -> Result<(), getrandom::Error> {
     RNG.with(|rng| {
         let buf = unsafe {
-            // fill the buffer with zeros
-            core::ptr::write_bytes(dest, 0, len);
             // create mutable byte slice
             core::slice::from_raw_parts_mut(dest, len)
         };
--- a/libs/orbit-essentials/src/utils/random.rs
+++ b/libs/orbit-essentials/src/utils/random.rs
@@ -19,22 +19,25 @@ thread_local! {
     target_vendor = "unknown",
     target_os = "unknown"
 ))]
+#[no_mangle]
 /// A getrandom implementation that works in the IC.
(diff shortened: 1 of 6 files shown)


libs/orbit-essentials/src/utils/random.rs
```diff
@@ -19,23 +19,26 @@ thread_local! {
...
 /// A getrandom implementation that works in the IC.
-pub fn custom_getrandom_bytes_impl(dest: &mut [u8]) -> Result<(), getrandom::Error> {
+unsafe extern "Rust" fn __getrandom_v03_custom(
+    dest: *mut u8,
+    len: usize,
+) -> Result<(), getrandom::Error> {
     RNG.with(|rng| {
+        let buf = unsafe {
+            // fill the buffer with zeros
+            core::ptr::write_bytes(dest, 0, len);
```
olaszakos, line 31: why do we need to zero it if we almost immediately overwrite it?

mraszyk: I was following [this](https://docs.rs/getrandom/0.3.1/getrandom/index.html#custom-backend) example. I guess the reason is to have predictable output in case the function called subsequently did not write all bytes.

olaszakos: I understand but that wouldn't be pretty problematic behaviour to not return the required random bytes, wouldn't it? 😅

olaszakos (approved):

libs/orbit-essentials/src/utils/random.rs
olaszakos: It's not about being more random, just that the fill_bytes should completely fill the buffer, otherwise this method's behavior is undefined. 

According to the docs, this behavior should be guaranteed: https://docs.rs/rand/latest/rand/trait.RngCore.html#tymethod.fill_bytes

If it fills all bytes then zeroing the buffer out beforehand is wasted instructions.

PR #581, opened 2025-05-27
chore(ci): bump PocketIC server to v9.0.3 and PocketIC library to v9.0.2

mraszyk:


mraszyk:
Keeping as draft since the PocketIC release used in this PR is broken on macOS.

PR #589, opened 2025-07-08
feat(station): request deduplication

tomer-dfinity:
This PR adds an optional deduplication key to a request. When a new request is created with a deduplication key its validation will fail if there's an existing request already in Created, Approved, Scheduled, or Processing states.
The deduplication key is stored as a Request index for faster lookup during validation.
The list_requests endpoint was modified so one could filter using the deduplication key.


core/station/impl/src/models/request.rs
```diff
@@ -177,6 +179,63 @@ fn validate_execution_plan(
...
+                    Request::MAX_DEDUPLICATION_KEY_LEN
+                ),
+            });
+        }
+
+        if !should_be_checked_for_duplicates {
+            return Ok(());
+        }
+
+        if deduplication_key.is_empty() {
```
keplervital, line 212: missing "trim" here as well

tomer-dfinity: Done

PR #597, opened 2025-08-29
fix(wallet): empty reason if no matching approval policy found

mraszyk:
This PR fixes the message "Request rejected for the following reasons: ." in case of no matching approval policy found to "Request rejected for the following reasons: No matching approval policy.".


mraszyk:
@keplervital I suppose there's some syntactic error in the code.

PR #634, opened 2026-05-22
feat(station): support wasm_memory_persistence for external canister upgrades

aterga:
## Summary

- `ChangeExternalCanisterOperationInput` gains two optional fields, `wasm_memory_persistence` and `skip_pre_upgrade`, plumbed through the API → internal model → `mgmt::CanisterInstallMode::Upgrade(Some(UpgradeFlags { … }))`.
- Without `wasm_memory_persistence = keep`, Motoko canisters that use Enhanced Orthogonal Persistence cannot be safely upgraded through Orbit — the IC defaults to `replace`, wiping their main memory.
- Fully backward-compatible: new fields are `opt`, legacy `CanisterUpgradeModeArgs` CBOR (empty map) still deserializes via `#[serde(default)]`.

## Motivation

The IC management canister's `install_code` accepts `CanisterUpgradeOptions` (containing `wasm_memory_persistence` and `skip_pre_upgrade`) when `mode` is `upgrade`. Until now, Orbit hard-coded this to `None`, which:

- Made it impossible to safely upgrade Motoko canisters using Enhanced Orthogonal Persistence — they require `wasm_memory_persistence = keep`, otherwise the IC clears their main memory on upgrade.
- Did not expose `skip_pre_upgrade`, useful for recovery scenarios where the existing `pre_upgrade` hook traps.

## What changed

- **Candid API** (`core/station/api`): new `WasmMemoryPersistence` variant; two new `opt` fields on `ChangeExternalCanisterOperationInput` and `ChangeExternalCanisterOperation`. `spec.did` and the wallet-generated `station.did{,.js,.d.ts}` mirror the additions.
- **Internal model** (`core/station/impl/src/models/request_operation.rs`): `CanisterUpgradeModeArgs` carries the two flag fields (with `#[serde(default)]` for backward-compat with legacy stored requests). `From<CanisterInstallMode>` now translates to `mgmt::CanisterInstallMode::Upgrade(Some(UpgradeFlags { … }))` when either field is set, or `Upgrade(None)` otherwise.
- **Mapper** (`core/station/impl/src/mappers/request_operation.rs`): the api↔impl conversions for the input and DTO route the flags through the upgrade variant.
- **dfx-orbit CLI** (`tools/dfx-orbit`): passes `None` for both fields when constructing requests (the CLI does not yet expose CLI flags for these — that can be a follow-up).
- **Tests**: five new unit tests covering mapper round-trip, mgmt translation, install/reinstall ignoring the flags, and legacy CBOR deserialization. The existing `check_candid_interface` test confirms `spec.did` matches the generated service.

## Test plan

- [x] `cargo test -p station --lib` — 388 unit tests pass, including new mapper tests and `check_candid_interface`.
- [x] `cargo test --workspace --lib --bins --exclude integration-tests` — 651 tests pass, 0 failures.
- [x] `cargo clippy --workspace --all-targets` — clean.
- [x] `cargo fmt --all -- --check` — clean.
- [ ] PocketIC integration tests — not run locally (no PocketIC binary present); CI will exercise them.
- [ ] Manual smoke test: upgrade a Motoko EOP canister via Orbit with `wasm_memory_persistence = Keep` and confirm stable state is preserved.

### DIFF
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -575,28 +575,18 @@ type RemoveUserGroupOperation = record {
 type CanisterInstallMode = variant {
   install;
   reinstall;
-  // Upgrade an existing canister. Carries optional flags that mirror the
-  // IC management canister's `CanisterUpgradeOptions`.
-  upgrade : opt CanisterUpgradeOptionsInput;
-};
-
-// Optional upgrade flags forwarded to the IC management canister's
-// `install_code` method when `mode` is `upgrade`.
-type CanisterUpgradeOptionsInput = record {
-  // Required as `keep` for upgrading Motoko canisters that use Enhanced
-  // Orthogonal Persistence; otherwise the IC clears their main memory.
-  wasm_memory_persistence : opt WasmMemoryPersistence;
-  // If `true`, the `pre_upgrade` hook is skipped during the canister upgrade.
-  skip_pre_upgrade : opt bool;
-};
-
-// WASM memory persistence setting passed to `install_code` when the install
-// mode is `upgrade`. `keep` is required for Motoko canisters that use
-// Enhanced Orthogonal Persistence; the IC defaults to `replace` (clearing
-// main memory) otherwise.
-type WasmMemoryPersistence = variant {
-  keep;
-  replace;
+  // Upgrade an existing canister. The optional record mirrors the IC
+  // management canister's `CanisterUpgradeOptions`.
+  // `wasm_memory_persistence = keep` is required for Motoko canisters that
+  // use Enhanced Orthogonal Persistence; otherwise the IC clears their main
+  // memory.
+  upgrade : opt record {
+    skip_pre_upgrade : opt bool;
+    wasm_memory_persistence : opt variant {
+      keep;
+      replace;
+    };
+  };
 };
 
 type SystemUpgradeTarget = variant {
--- a/core/station/api/spec.did
+++ b/core/station/api/spec.did
@@ -575,7 +575,19 @@ type RemoveUserGroupOperation = record {
 type CanisterInstallMode = variant {
   install;
   reinstall;
-  upgrade;
+  // Upgrade an existing canister. Carries optional flags that mirror the
+  // IC management canister's `CanisterUpgradeOptions`.
+  upgrade : opt CanisterUpgradeOptionsInput;
+};
+
+// Optional upgrade flags forwarded to the IC management canister's
+// `install_code` method when `mode` is `upgrade`.
+type CanisterUpgradeOptionsInput = record {
+  // Required as `keep` for upgrading Motoko canisters that use Enhanced
+  // Orthogonal Persistence; otherwise the IC clears their main memory.
+  wasm_memory_persistence : opt WasmMemoryPersistence;
+  // If `true`, the `pre_upgrade` hook is skipped during the canister upgrade.
+  skip_pre_upgrade : opt bool;
 };
 
 // WASM memory persistence setting passed to `install_code` when the install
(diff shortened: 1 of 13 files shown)


core/station/api/spec.did
```diff
@@ -666,6 +675,13 @@ type ChangeExternalCanisterOperationInput = record {
   module_extra_chunks : opt WasmModuleExtraChunks;
   // The initial argument passed to the new wasm module.
   arg : opt blob;
+  // WASM memory persistence setting. Only applicable when `mode` is `upgrade`.
+  // Required as `keep` for upgrading Motoko canisters that use Enhanced
+  // Orthogonal Persistence; otherwise the IC clears their main memory.
+  wasm_memory_persistence : opt WasmMemoryPersistence;
+  // If `true`, the `pre_upgrade` hook is skipped. Only applicable when `mode`
+  // is `upgrade`.
+  skip_pre_upgrade : opt bool;
```
mraszyk: I'd recommend to extend the upgrade variant of `CanisterInstallMode` with those fields.

mraszyk: For compatibility reasons, the new type should be:
```
    upgrade : opt record {
        skip_pre_upgrade : opt bool;
        wasm_memory_persistence : opt variant {
            keep;
            replace;
        };
    };
```

aterga: Done in 25514cac — moved the flags into `CanisterInstallMode::Upgrade(opt CanisterUpgradeOptionsInput)` and dropped the flat fields from the operation input/DTO.

aterga: Done in 9161346a — inlined the record and variant per your suggestion, and regenerated the wallet TS/JS bindings. Rust types stay named (Rust enum variants can't carry anonymous structs); `service_equal` confirms structural equivalence with the inline spec.

core/station/impl/src/models/request_operation.rs
```diff
@@ -476,7 +497,18 @@ impl From<CanisterInstallMode> for mgmt::CanisterInstallMode {
...
+            CanisterInstallMode::Upgrade(args) => {
+                let flags =
+                    if args.wasm_memory_persistence.is_some() || args.skip_pre_upgrade.is_some() {
+                        Some(mgmt::UpgradeFlags {
+                            skip_pre_upgrade: args.skip_pre_upgrade,
+                            wasm_memory_persistence: args.wasm_memory_persistence.map(Into::into),
+                        })
+                    } else {
+                        None
+                    };
```
mraszyk: Instead of this ad-hoc conversion, I'll suggest to implement From for CanisterUpgradeModeArgs and map over it

core/station/impl/src/mappers/request_operation.rs
```diff
@@ -552,13 +561,46 @@ impl From<CanisterInstallMode> for station_api::CanisterInstallMode {
...
             }
-            CanisterInstallMode::Upgrade(CanisterUpgradeModeArgs {}) => {
-                station_api::CanisterInstallMode::Upgrade
+            CanisterInstallMode::Upgrade(args) => {
+                let opts =
+                    if args.wasm_memory_persistence.is_some() || args.skip_pre_upgrade.is_some() {
+                        Some(args.into())
+                    } else {
+                        None
+                    };
```
mraszyk: let's map over Into::into here, too

mraszyk (approved):

core/station/impl/src/models/request_operation.rs
aterga: Done in 24382a3e — widened internal `Upgrade` to `Option<CanisterUpgradeModeArgs>` and added `From<CanisterUpgradeModeArgs> for mgmt::UpgradeFlags`; the conversion is now just `args.map(Into::into)`.

core/station/impl/src/mappers/request_operation.rs
aterga: Done in 24382a3e — same change here, the impl→api mapper is now `args.map(Into::into)` over `Option<CanisterUpgradeModeArgs>`.

sea-snake (approved):

PR #638, opened 2026-06-24
feat(cli): audit subcommand for station configuration sanity checks

MRmarioruci:
## Summary

- New `orbit-cli audit --station <CANISTER_ID>` subcommand: read-only sanity checks against an Orbit station's configuration.
- Three checks ship in this first release: an empty-approver-set Quorum trap, a validation/execution-method match on `CallExternalCanister`, and an `EditAsset`-weaker-than-`Transfer` configuration warning.
- Uses the same `dfx canister call --output json` pattern as `registry/` and `release/`. No new runtime dependencies. Read-only by construction — only query methods are invoked.

## Motivation

Several station-configuration shapes that look fine in the wallet UI can silently disable approval gates or create routing-redirect vectors:

- A `Quorum` / `QuorumPercentage` rule whose `UserSpecifier` resolves to zero active users auto-approves with no votes — the evaluator's `min(min_approved, total_possible_approvers)` clamp drives the threshold to `0`.
- A `CallExternalCanister` permission or policy where `validation_method == execution_method` runs the side effect at submission time, before the approval policy has completed.
- An `EditAsset` policy gated more loosely than `Transfer` for the same asset lets an actor mutate `asset.metadata["ledger_canister_id"]` between an approved transfer's policy evaluation and its batch-job execution.

These all require manual UI walkthroughs to spot today. `orbit-cli audit` makes the check a one-command read-only sweep that can run on demand or in CI.

## What changed

- **New** `cli/src/audit/`:
  - `index.ts` — Commander command (`-s/--station`, `-n/--network`, `-i/--identity`, `-o/--output`).
  - `station.core.ts` — paginated `dfx canister call` wrappers for `list_request_policies`, `list_users`, `list_user_groups`, `list_assets`, `list_named_rules`, `list_permissions`.
  - `types.ts` — hand-written TS subset of the station API; small enough that the `didc bind` toolchain dependency isn't worth pulling in for this MVP.
  - `resolver.ts` — `resolveApprovers` and `minVotesForRule` mirroring the relevant subset of `find_matching_users` and `RequestApprovalSummary::evaluate` from `core/station/impl/src/models/request_policy_rule.rs`; walks `NamedRule` chains with cycle detection.
  - `report.ts` — `Finding` type, severity-sorted text formatter, exit-code helper.
  - `checks/*.ts` — one file per check; each returns `Finding[]`.
  - `checks/*.spec.ts` — unit tests, 19 cases across positive and negative paths.
  - `README.md` — usage, exit codes, check catalogue, extension notes, roadmap.
- **`cli/src/cli.ts`** — register the new subcommand alongside `registry` and `release`.
- **`cli/package.json`** — add `vitest 1.6.1` as a dev dependency (matching the root pin) and a `test` script.
- **`pnpm-lock.yaml`** — updated for the vitest addition.

Exit codes: `0` clean / `1` warnings only / `2` any blockers. Convenient for CI gating.

`--output <PATH>` writes the report to a file (overwrite); the exit code is unchanged and a short confirmation is emitted to stderr so file-mode runs still give feedback without polluting the report.

## Test plan

- [x] `pnpm --filter orbit-cli test` — 19 vitest cases pass (3 spec files). Covers each check's positive and negative paths, combinator descent (`AnyOf`), `NamedRule` resolution, and cycle detection.
- [x] `pnpm --filter orbit-cli build` — `tsc` clean.
- [x] Prettier + ESLint clean on all new files.
- [x] Manual smoke test against a mainnet station: created an empty user group, a NamedRule pointing at it, a Request Policy referencing the NamedRule, a `CallExternalCanister` permission with `validation == execution`, and a weak `EditAsset` policy — the audit reported 2 BLOCKERs + 1 WARNING with exit code 2. Cleanup returned the station to 0 findings, exit code 0.
- [ ] Reviewers may want to run it against their own local or mainnet stations.

## Notes for reviewers

- The implementation deliberately mirrors the existing `registry/` subcommand's structure (shell out via `dfx canister call --output json`, paginated `list_*`, hand-typed result shapes) so it slots into the codebase without introducing new patterns.
- The `asset.edit-policy-weaker-than-transfer` check compares station-wide rather than per-asset/per-account — that's a deliberate MVP shortcut documented in the README's roadmap. Per-asset scoping is a planned follow-up.
- No security-ticket / disclosure references appear in the code or commit history; the README explains each check in technical terms only.

### DIFF
--- a/cli/src/audit/checks/asset-edit-policy-weaker-than-transfer.ts
+++ b/cli/src/audit/checks/asset-edit-policy-weaker-than-transfer.ts
@@ -42,7 +42,7 @@ export const assetEditPolicyWeakerThanTransfer = (
       checkId: 'asset.edit-policy-weaker-than-transfer',
       severity: 'warning',
       location: `station-wide (${editPolicies.length} EditAsset policy/policies vs ${transferPolicies.length} Transfer policy/policies)`,
-      message: `Easiest EditAsset path requires ${easiestEdit} approval(s); strictest Transfer path requires ${strictestTransfer}. An attacker who can pass the EditAsset bar can mutate asset.metadata["ledger_canister_id"] and redirect an approved Transfer between approval and execution.`,
+      message: `Easiest EditAsset path requires at least ${easiestEdit} approval(s); strictest Transfer path requires at least ${strictestTransfer} (estimates — AllOf combinators are reduced via max-of-children, giving a lower bound). An actor who can pass the EditAsset bar can mutate asset.metadata["ledger_canister_id"] and redirect an approved Transfer between approval and execution.`,
       fix: 'Gate EditAsset at the same approval level as Transfer for the affected assets. Until v0.2 supports per-asset scoping, treat this as a station-wide signal to tighten the loosest EditAsset policy.',
     },
   ];
--- a/cli/src/audit/checks/asset-edit-policy-weaker-than-transfer.ts
+++ b/cli/src/audit/checks/asset-edit-policy-weaker-than-transfer.ts
@@ -1,0 +1,49 @@
+import { Finding } from '../report';
+import { minVotesForRule } from '../resolver';
+import { NamedRule, RequestPolicy, User, UUID } from '../types';
+
+/**
+ * Warns when the *easiest path* to passing an `EditAsset` request requires
+ * fewer approvals than the *strictest* `Transfer` policy in the station.
+ *
+ * Background: an approved transfer's destination ledger is resolved live from
+ * `asset.metadata["ledger_canister_id"]` at execute time. A successful
+ * `EditAsset` between approval and execution can therefore redirect funds. If
+ * `EditAsset` is gated more loosely than `Transfer`, an attacker who can pass
+ * the lower bar can subvert the higher one.
+ *
+ * Per-asset / per-account scoping would be more precise; the MVP compares the
+ * easiest EditAsset path against the strictest Transfer path across the entire
+ * station, which catches the typical misconfiguration without requiring an
+ * account-to-asset cross-reference.
+ */
+export const assetEditPolicyWeakerThanTransfer = (
+  policies: RequestPolicy[],
+  users: User[],
+  namedRules: NamedRule[],
+): Finding[] => {
+  const namedByUUID = new Map<UUID, NamedRule>(namedRules.map(r => [r.id, r]));
+  const editPolicies = policies.filter(p => 'EditAsset' in p.specifier);
+  const transferPolicies = policies.filter(p => 'Transfer' in p.specifier);
+
+  if (editPolicies.length === 0 || transferPolicies.length === 0) return [];
+
+  const editVotes = editPolicies.map(p => minVotesForRule(p.rule, users, namedByUUID));
+  const transferVotes = transferPolicies.map(p => minVotesForRule(p.rule, users, namedByUUID));
+
+  const easiestEdit = Math.min(...editVotes);
+  const strictestTransfer = Math.max(...transferVotes);
+
+  if (!Number.isFinite(easiestEdit) || !Number.isFinite(strictestTransfer)) return [];
+  if (easiestEdit >= strictestTransfer) return [];
+
+  return [
+    {
+      checkId: 'asset.edit-policy-weaker-than-transfer',
+      severity: 'warning',
+      location: `station-wide (${editPolicies.length} EditAsset policy/policies vs ${transferPolicies.length} Transfer policy/policies)`,
+      message: `Easiest EditAsset path requires ${easiestEdit} approval(s); strictest Transfer path requires ${strictestTransfer}. An attacker who can pass the EditAsset bar can mutate asset.metadata["ledger_canister_id"] and redirect an approved Transfer between approval and execution.`,
+      fix: 'Gate EditAsset at the same approval level as Transfer for the affected assets. Until v0.2 supports per-asset scoping, treat this as a station-wide signal to tighten the loosest EditAsset policy.',
+    },
+  ];
+};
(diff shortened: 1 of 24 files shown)


cli/src/audit/station.core.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const PAGE_SIZE = 50;
+
+const dfx = async (ctx: StationContext, method: string, args: string): Promise<unknown> => {
+  // Candid arguments are always tuples — even single-arg calls need the outer `( ... )`.
+  const cmd = `dfx canister call --identity '${ctx.identity}' --network '${ctx.network}' --output json '${ctx.station}' ${method} '(${args})'`;
+  const raw = await execAsync(cmd);
+  return JSON.parse(raw);
+};
```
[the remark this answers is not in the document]
MRmarioruci: Good catch — fixed in 3eab3299. Added a `shq()` helper that wraps each user-controlled value (`identity` / `network` / `station`) in single-quoted POSIX shell form with embedded quotes escaped via the standard `'` idiom. `method` and `args` are constructed in this file from a closed set of values, so they don't need escaping. Switching to `execFile` with an argv array would be cleaner; left it as escaping for now since `registry/` and `release/` use the same `dfx canister call ...` shell pattern, and a tightening pass across all three is better done together as a follow-up.

cli/src/audit/checks/quorum-empty-approver-set.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  for (const policy of policies) {
+    walkQuorumRules(policy.rule, namedByUUID, (kind, approvers, minApproved, path) => {
+      const resolved = resolveApprovers(approvers, users);
+      if (resolved.length === 0 && minApproved > 0) {
+        findings.push({
+          checkId: 'quorum.empty-approver-set',
+          severity: 'blocker',
+          location: `policy ${policy.id} (${describeRequestSpecifier(policy.specifier)}) — ${path}`,
+          message: `${kind} rule asks for ${minApproved} approval(s) but ${describeSpecifier(approvers)} currently resolves to 0 active users. Next matching request will auto-approve.`,
```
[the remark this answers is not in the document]
MRmarioruci: Right — fixed in 3eab3299. The check now renders QuorumPercentage findings as `"50% of approvers"` and Quorum as `"N approval(s)"`. Added a positive test (`renders QuorumPercentage findings as a percentage, not a vote count`) to lock the wording.

cli/src/audit/checks/asset-edit-policy-weaker-than-transfer.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  if (!Number.isFinite(easiestEdit) || !Number.isFinite(strictestTransfer)) return [];
+  if (easiestEdit >= strictestTransfer) return [];
+
+  return [
+    {
+      checkId: 'asset.edit-policy-weaker-than-transfer',
+      severity: 'warning',
+      location: `station-wide (${editPolicies.length} EditAsset policy/policies vs ${transferPolicies.length} Transfer policy/policies)`,
+      message: `Easiest EditAsset path requires ${easiestEdit} approval(s); strictest Transfer path requires ${strictestTransfer}. An attacker who can pass the EditAsset bar can mutate asset.metadata["ledger_canister_id"] and redirect an approved Transfer between approval and execution.`,
```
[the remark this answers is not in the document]
MRmarioruci: Fair — the finding message now says `"requires at least N approval(s)"` and notes parenthetically that the estimate is a lower bound (AllOf reduced via max-of-children). That keeps the warning honest without losing the comparison the operator needs. Fixed in 3eab3299.

cli/src/audit/station.core.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+const PAGE_SIZE = 50;
+
+// Escapes a value for safe inclusion in a single-quoted POSIX shell argument.
+// A literal `'` inside single quotes is impossible, so the standard idiom is to
+// close the string, emit an escaped quote, and reopen: foo'bar -> 'foo'\''bar'.
+const shq = (value: string): string => `'${value.replace(/'/g, "'\\''")}'`;
+
+const dfx = async (ctx: StationContext, method: string, args: string): Promise<unknown> => {
```
aterga: Could we avoid using DFX here whatsoever? 

In theory we could use agent-js (preferable approach; might require more work)

Ot at least we could use ICP CLI.

NCR if this is a non-trivial change for the PR.

cli/src/audit/resolver.ts
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * Combinators reduce as:
+ *   AnyOf  → min over children (easiest path)
+ *   AllOf  → max over children (lower-bound; semantically each child needs its own quorum)
+ *   Not    → opaque, treated as Infinity (cannot reason statically about negation)
+ *
+ * `AutoApproved` and `AllowListed*` return 0 — rules that bypass vote counting.
+ * Unresolvable cases (cycle, missing NamedRule) return Infinity to avoid false positives.
+ */
+export const minVotesForRule = (
```
aterga, line 105: Are there code-level specs for this function?

cli/src/audit/station.core.ts
MRmarioruci: we can use agent-js. will adjust

cli/src/audit/resolver.ts
MRmarioruci: Good call — only had indirect coverage through the asset-policy check. Added [`cli/src/audit/resolver.spec.ts`](https://github.com/dfinity/orbit/blob/66ee6846/cli/src/audit/resolver.spec.ts) in 66ee6846 with 32 direct unit tests across `resolveApprovers`, `walkQuorumRules`, and `minVotesForRule`.

aterga (approved):

cli/src/audit/station.core.ts
MRmarioruci: Done, swapped to `@dfinity/agent` in 3c673f85.

PR #640, opened 2026-06-24
docs(cli): add orbit-station-audit Claude Code skill

aterga:
## What

Adds a **project skill** at `.claude/skills/orbit-station-audit/` that takes an agent from a fresh checkout to a successful `orbit-cli audit` run against a live station.

```
.claude/skills/orbit-station-audit/
├── SKILL.md              # operational guide
└── scripts/run-audit.sh  # build-from-source → repair generated IDL → run audit
```

## Why

The existing [`cli/src/audit/README.md`](../blob/main/cli/src/audit/README.md) documents flags, exit codes, and report format well — but a first run still trips on operational details that aren't in the README. This skill front-loads exactly those, and points to the README for reference rather than duplicating it.

The four gotchas it captures (all hit live while writing it):

1. **The global `orbit-cli` is unreliable** — it either predates the `audit` subcommand, or (after a `pnpm install`, which rebuilds + `pnpm link --global` re-exposes via the `prepare-cli` postinstall) carries the build bug below and crashes. Build and run the repo's `cli/dist/cli.js` directly.
2. **The built CLI crashes with `SyntaxError: Unexpected token 'export'`** — the `build` script copies the generated `station.did.js` (an ES module) into the CommonJS bundle, so `require()` can't parse it. The helper script repairs the dist copy after each build. _The real fix belongs in the CLI build itself — happy to follow up with that separately._
3. **Internet Identity principals are per-origin** — the audit must be called by a principal the station knows as a member, so the `icp-cli` identity must be derived from `orbitwallet.io` (the `derivationOrigin` pinned in `apps/wallet/src/configs/init.config.ts`), passed as a bare host to `--app`.
4. **`icp identity link web` waits for an Enter keypress** before opening the browser — pipe `printf '\n'` when running it non-interactively.

## Verification

Built from a deliberately re-broken (clean) state, the helper script repaired the IDL and ran the audit end-to-end against a mainnet station, producing a severity-sorted report (exit code 2). The content was checked by a three-lens review (cold-execution, conciseness, factual-accuracy) against the codebase; all load-bearing claims verified.

### DIFF
--- a/.claude/skills/orbit-station-audit/scripts/run-audit.sh
+++ b/.claude/skills/orbit-station-audit/scripts/run-audit.sh
@@ -37,10 +37,18 @@ fi
 #
 #    The proper fix is to emit/copy this file as CommonJS in the CLI's build
 #    script (cli/package.json) so this guard becomes unnecessary.
-if grep -q '^export const' "$GEN"; then
-  perl -0pi -e 's/^export const /const /mg' "$GEN"
-  printf '\nmodule.exports = { idlFactory, init };\n' >>"$GEN"
-fi
+#
+#    Done in Node (already a hard requirement here) so the script needs no extra
+#    runtime like perl/sed. Idempotent: a no-op once the file is already CJS.
+node -e '
+  const fs = require("fs");
+  const p = process.argv[1];
+  let s = fs.readFileSync(p, "utf8");
+  if (s.includes("export const ")) {
+    s = s.replace(/^export const /gm, "const ") + "\nmodule.exports = { idlFactory, init };\n";
+    fs.writeFileSync(p, s);
+  }
+' "$GEN"
 
 # 4. Run the audit, forwarding all arguments (default to --help if none given,
 #    since `audit` requires --station and would otherwise error).
--- a/.claude/skills/orbit-station-audit/scripts/run-audit.sh
+++ b/.claude/skills/orbit-station-audit/scripts/run-audit.sh
@@ -2,9 +2,9 @@
 #
 # Build the Orbit CLI from source and run `orbit-cli audit`.
 #
-# Why build from source instead of the global `orbit-cli`: the `audit`
-# subcommand is recent, and a globally-installed orbit-cli is almost certainly
-# older and doesn't have it. Building here keeps us on the checked-out version.
+# Why build from source instead of the global `orbit-cli`: a global on PATH may
+# be from an older checkout that predates the `audit` subcommand. Building here
+# keeps us on the version in this repo, with no doubt about what's invoked.
 #
 # Usage:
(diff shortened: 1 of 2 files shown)


.claude/skills/orbit-station-audit/scripts/run-audit.sh
```diff
@@ -0,0 +1,48 @@
...
+#    CommonJS — so `require()` throws `Unexpected token 'export'`. We rewrite the
+#    dist copy to CommonJS. This runs after every build because the copy step
+#    re-introduces the ESM file each time.
+#
+#    The proper fix is to emit/copy this file as CommonJS in the CLI's build
+#    script (cli/package.json) so this guard becomes unnecessary.
+if grep -q '^export const' "$GEN"; then
+  perl -0pi -e 's/^export const /const /mg' "$GEN"
+  printf '\nmodule.exports = { idlFactory, init };\n' >>"$GEN"
+fi
```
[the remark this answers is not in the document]
aterga: Good call — swapped the `perl` rewrite for a self-contained, idempotent Node snippet (Node is already a hard requirement here), so the script no longer depends on perl. Verified end-to-end: a re-broken `dist` is repaired and `orbit-cli audit` loads.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.claude/skills/orbit-station-audit/scripts/run-audit.sh
```diff
@@ -0,0 +1,56 @@
...
+#    CommonJS — so `require()` throws `Unexpected token 'export'`. We rewrite the
+#    dist copy to CommonJS. This runs after every build because the copy step
+#    re-introduces the ESM file each time.
+#
+#    The proper fix is to emit/copy this file as CommonJS in the CLI's build
+#    script (cli/package.json) so this guard becomes unnecessary.
+#
+#    Done in Node (already a hard requirement here) so the script needs no extra
+#    runtime like perl/sed. Idempotent: a no-op once the file is already CJS.
+node -e '
```
MRmarioruci: This should not be an issue anymore I think after the previous PR landed

MRmarioruci (approved):

.claude/skills/orbit-station-audit/scripts/run-audit.sh
aterga: You're right — #639 emits the IDL factory as CommonJS, so the crash is gone. Merged main and confirmed `pnpm --filter orbit-cli build` now runs `audit` directly with no repair. Removed the repair step from this script and dropped the ESM gotcha + troubleshooting entry from SKILL.md (four gotchas → three).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #641, opened 2026-07-01
feat(dfx-orbit,wallet): support --wasm-memory-persistence for external canister upgrades

aterga:
## Summary

The station already accepts `wasm_memory_persistence` and `skip_pre_upgrade` on external-canister upgrades (via `CanisterUpgradeOptionsInput`, added in #634), but neither of the two places users actually trigger an upgrade exposed them. This PR wires those upgrade options through the **dfx-orbit CLI** and the **wallet Install screen**.

Without `wasm_memory_persistence = keep`, Motoko canisters that use [Enhanced Orthogonal Persistence](https://docs.internetcomputer.org/motoko/orthogonal-persistence/enhanced) cannot be safely upgraded through Orbit — the IC defaults to `replace`, wiping their main memory.

### dfx-orbit CLI (`tools/dfx-orbit`)
- New flags on `request canister install` / `verify canister install`:
  - `--wasm-memory-persistence keep|replace`
  - `--skip-pre-upgrade`
- The flags are folded into `CanisterInstallMode::Upgrade(Some(..))` and are **rejected for `install`/`reinstall`** (the IC only honours them on upgrades).
- `verify` now compares the **full** install mode (including the upgrade options), so a request whose stored persistence mode differs from what the verifier expects is rejected instead of silently passing.
- The change-canister request display now prints the upgrade options.
- README documents the new flags.

### Wallet Install screen (`apps/wallet`)
- New **"Wasm Memory Persistence"** select (Default / Keep / Replace) and a **"Skip pre-upgrade hook"** toggle that appear only when the install mode is `upgrade`, wired into the install model's `mode`.
- When neither option is set the request collapses back to a plain `{ upgrade: [] }`, keeping it identical to the previous behaviour.
- New `CanisterWasmMemoryPersistenceSelect` input component and locale strings (en/fr/pt).

### API
- Derive `PartialEq`/`Eq` for `station_api::CanisterInstallMode` so the CLI can compare the full mode during verification (additive; does not change the candid interface).

## Tests
- **dfx-orbit unit tests**: `install_mode()` folds each flag correctly, produces `Upgrade(None)` when unset, and rejects the flags for install/reinstall with the exact error message.
- **Integration test**: an upgrade request carrying `wasm_memory_persistence = replace` + `skip_pre_upgrade` round-trips through the station; `verify` accepts a matching request and rejects a mismatched persistence mode; the approved upgrade then executes, proving the options are forwarded to the management canister. (`keep` cannot execute against the non-EOP test wasm — the IC rejects it for modules without the EOP metadata section.)
- **Wallet component tests**: the upgrade options are hidden unless the mode is `upgrade`, both options fold into `mode`, clearing them collapses back to a plain upgrade, and the persistence select's option↔candid-union mapping is covered by a dedicated spec.

## Verification performed here
- `wallet` `vue-tsc --noEmit` type-check: clean.
- `wallet` vitest for the touched components/inputs: passing.
- `prettier --check` and `eslint` on the changed wallet files: clean.
- `rustfmt --check` on the changed Rust files: clean.
- ⚠️ A full `cargo build`/`cargo test`/`clippy` could not run in this environment because the `dfinity/response-verification` git dependency is outside the sandbox's egress allow-list. CI exercises the Rust build, unit tests, and PocketIC integration tests.

### DIFF
--- a/apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
+++ b/apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
@@ -64,8 +64,11 @@ const selected = computed<WasmMemoryPersistenceOption>({
       case 'replace':
         emit('update:modelValue', { replace: null });
         break;
-      default:
+      case 'default':
+        // `default` means "let the IC decide" and maps to an absent
+        // wasm_memory_persistence option in the request.
         emit('update:modelValue', undefined);
+        break;
     }
   },
 });
--- a/apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
+++ b/apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
@@ -1,0 +1,78 @@
+<template>
+  <VSelect
+    v-model="selected"
+    :label="label"
+    :variant="props.variant"
+    :density="props.density"
+    :readonly="props.readonly"
+    :items="items"
+    :hint="props.hint"
+    :persistent-hint="props.hint !== undefined"
+    :prepend-icon="mdiMemory"
+  />
+</template>
+<script setup lang="ts">
+import { mdiMemory } from '@mdi/js';
+import { computed } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { VSelect } from 'vuetify/components';
+import { WasmMemoryPersistence } from '~/components/external-canisters/external-canisters.types';
+
+type WasmMemoryPersistenceOption = 'default' | 'keep' | 'replace';
+
+const props = withDefaults(
+  defineProps<{
+    modelValue?: WasmMemoryPersistence;
+    readonly?: boolean;
+    label?: string;
+    hint?: string;
+    density?: 'comfortable' | 'compact' | 'default';
+    variant?: 'filled' | 'outlined' | 'plain' | 'solo' | 'underlined';
+  }>(),
+  {
+    modelValue: undefined,
+    readonly: false,
+    label: undefined,
+    hint: undefined,
+    density: 'comfortable',
+    variant: 'filled',
+  },
+);
+
+const emit = defineEmits<{
+  (event: 'update:modelValue', payload?: WasmMemoryPersistence): void;
+}>();
+
+const i18n = useI18n();
+const label = computed(
+  () => props.label ?? i18n.t('external_canisters.wasm_memory_persistence.label'),
+);
+
+const selected = computed<WasmMemoryPersistenceOption>({
+  get: () => {
+    if (!props.modelValue) {
+      return 'default';
+    }
+
+    return 'keep' in props.modelValue ? 'keep' : 'replace';
+  },
+  set: value => {
+    switch (value) {
+      case 'keep':
+        emit('update:modelValue', { keep: null });
+        break;
+      case 'replace':
+        emit('update:modelValue', { replace: null });
+        break;
+      default:
+        emit('update:modelValue', undefined);
+    }
+  },
+});
+
+const items = computed<{ title: string; value: WasmMemoryPersistenceOption }[]>(() => [
+  { title: i18n.t('external_canisters.wasm_memory_persistence.default'), value: 'default' },
+  { title: i18n.t('external_canisters.wasm_memory_persistence.keep'), value: 'keep' },
+  { title: i18n.t('external_canisters.wasm_memory_persistence.replace'), value: 'replace' },
+]);
+</script>
--- a/tests/integration/src/dfx_orbit/install.rs
+++ b/tests/integration/src/dfx_orbit/install.rs
@@ -8,7 +8,9 @@ use crate::{
     TestEnv,
 };
 use candid::Encode;
-use dfx_orbit::canister::{CanisterInstallModeArgs, RequestCanisterInstallArgs};
+use dfx_orbit::canister::{
+    CanisterInstallModeArgs, RequestCanisterInstallArgs, WasmMemoryPersistenceArgs,
+};
 use dfx_orbit::{
     args::{RequestArgs, RequestArgsActions, VerifyArgs, VerifyArgsAction},
     canister::{
@@ -85,6 +87,8 @@ fn canister_install(use_chunks: bool) {
     let inner_args = RequestCanisterInstallArgs {
         canister: String::from("test"),
         mode: CanisterInstallModeArgs::Install,
+        wasm_memory_persistence: None,
+        skip_pre_upgrade: false,
         wasm: wasm.path().as_os_str().to_str().unwrap().to_string(),
         argument: None,
         arg_file: None,
@@ -151,4 +155,108 @@ fn canister_install(use_chunks: bool) {
     // Check that the canister is installed now.
     let status = canister_status(&env, Some(canister_ids.station), test_canister);
     assert_eq!(status.module_hash, Some(module_hash));
+}
+
+/// Test that `--wasm-memory-persistence` and `--skip-pre-upgrade` are plumbed
+/// through into the upgrade request and survive the round-trip through the
+/// station, so that `verify` accepts a matching request and rejects a
+/// mismatched one. The request is left pending (four-eyes is enabled and only
+/// the requester has approved) so the incompatible `keep` upgrade never
+/// executes against the empty test canister.
+#[test]
+fn canister_upgrade_options_round_trip() {
+    let TestEnv {
+        mut env,
+        canister_ids,
+        ..
+    } = setup_new_env();
+
+    let (_dfx_user, _) = setup_dfx_user(&env, &canister_ids);
+    let other_user = user_test_id(1);
+    add_user(&env, other_user, vec![], canister_ids.station);
+
+    let test_canister = create_canister(&env, canister_ids.station);
+
+    permit_change_operation(&env, &canister_ids);
+    set_four_eyes_on_change(&env, &canister_ids);
+
+    let config = DfxOrbitTestConfig {
+        canister_ids: vec![(String::from("test"), test_canister)],
+        ..Default::default()
+    };
+
+    let mut wasm = NamedTempFile::new().unwrap();
+    let module_bytes = get_canister_wasm("test_canister");
+    wasm.write_all(&module_bytes).unwrap();
+
+    let inner_args = RequestCanisterInstallArgs {
+        canister: String::from("test"),
+        mode: CanisterInstallModeArgs::Upgrade,
+        wasm_memory_persistence: Some(WasmMemoryPersistenceArgs::Keep),
+        skip_pre_upgrade: true,
+        wasm: wasm.path().as_os_str().to_str().unwrap().to_string(),
+        argument: None,
+        arg_file: None,
+        asset_canister: None,
+    };
+
+    dfx_orbit_test(&mut env, config, async {
+        let dfx_orbit = setup_dfx_orbit(canister_ids.station).await;
+
+        let request = RequestArgs {
+            title: None,
+            summary: None,
+            action: RequestArgsActions::Canister(RequestCanisterArgs {
+                action: RequestCanisterActionArgs::Install(inner_args.clone()),
+            }),
+        }
+        .into_request(&dfx_orbit)
+        .await
+        .unwrap();
+
+        let request = dfx_orbit.station.request(request.clone()).await.unwrap();
+
+        let req_response = dfx_orbit
+            .station
+            .review_id(GetRequestInput {
+                request_id: request.request.id.clone(),
+                with_full_info: Some(false),
+            })
+            .await
+            .unwrap();
+
+        // The stored upgrade options match, so verification succeeds.
+        VerifyArgs {
+            request_id: request.request.id.clone(),
+            and_approve: false,
+            or_reject: false,
+            action: VerifyArgsAction::Canister(VerifyCanisterArgs {
+                action: VerifyCanisterActionArgs::Install(inner_args.clone()),
+            }),
+        }
+        .verify(&dfx_orbit, &req_response)
+        .await
+        .unwrap();
+
+        // A mismatched `wasm_memory_persistence` must fail verification.
+        let mut mismatched_args = inner_args.clone();
+        mismatched_args.wasm_memory_persistence = Some(WasmMemoryPersistenceArgs::Replace);
+        VerifyArgs {
+            request_id: request.request.id.clone(),
+            and_approve: false,
+            or_reject: false,
+            action: VerifyArgsAction::Canister(VerifyCanisterArgs {
+                action: VerifyCanisterActionArgs::Install(mismatched_args),
+            }),
+        }
+        .verify(&dfx_orbit, &req_response)
+        .await
+        .expect_err("verification must reject a mismatched wasm_memory_persistence");
+
+        request.request
+    });
+
+    // The upgrade request is still pending, so the canister remains empty.
+    let status = canister_status(&env, Some(canister_ids.station), test_canister);
+    assert_eq!(status.module_hash, None);
 }
--- a/tests/integration/src/dfx_orbit/install.rs
+++ b/tests/integration/src/dfx_orbit/install.rs
@@ -20,7 +20,7 @@ use dfx_orbit::{
 };
 use flate2::read::GzDecoder;
 use sha2::{Digest, Sha256};
-use station_api::{GetRequestInput, RequestApprovalStatusDTO};
+use station_api::{GetRequestInput, RequestApprovalStatusDTO, RequestStatusDTO};
 use std::io::{Read, Write};
 use tempfile::NamedTempFile;
 
@@ -382,9 +382,9 @@ fn canister_upgrade_with_wasm_memory_persistence_keep() {
 
     let (request_without_keep, request_with_keep) = dfx_orbit_test(&mut env, config, async {
         let dfx_orbit = setup_dfx_orbit(canister_ids.station).await;
+        let dfx_orbit = &dfx_orbit;
 
-        let mut requests = Vec::new();
-        for args in [args_without_keep, args_with_keep] {
+        let submit_upgrade = move |args: RequestCanisterInstallArgs| async move {
             let request = RequestArgs {
                 title: None,
                 summary: None,
@@ -392,21 +392,23 @@ fn canister_upgrade_with_wasm_memory_persistence_keep() {
                     action: RequestCanisterActionArgs::Install(args),
                 }),
             }
-            .into_request(&dfx_orbit)
+            .into_request(dfx_orbit)
             .await
             .unwrap();
 
-            let response = dfx_orbit.station.request(request).await.unwrap();
-            requests.push(response.request);
-        }
+            dfx_orbit.station.request(request).await.unwrap().request
+        };
 
-        let with_keep = requests.pop().unwrap();
-        let without_keep = requests.pop().unwrap();
-        (without_keep, with_keep)
+        let request_without_keep = submit_upgrade(args_without_keep).await;
+        let request_with_keep = submit_upgrade(args_with_keep).await;
+        (request_without_keep, request_with_keep)
     });
 
     // Without `keep`, the IC refuses to drop the main memory of a canister
-    // running an Enhanced Orthogonal Persistence module, so the upgrade fails.
+    // running an Enhanced Orthogonal Persistence module, so the upgrade fails
+    // during execution. Confirm it failed for that reason — the `install_code`
+    // call was rejected — rather than being rejected/cancelled at an earlier
+    // stage or timing out.
     submit_request_approval(
         &env,
         other_user,
@@ -414,8 +416,25 @@ fn canister_upgrade_with_wasm_memory_persistence_keep() {
         request_without_keep.clone(),
         RequestApprovalStatusDTO::Approved,
     );
-    wait_for_request(&env, other_user, canister_ids.station, request_without_keep)
-        .expect_err("the IC must reject an EOP upgrade without wasm_memory_persistence = keep");
+    let failed_status =
+        wait_for_request(&env, other_user, canister_ids.station, request_without_keep)
+            .expect_err("the IC must reject an EOP upgrade without wasm_memory_persistence = keep")
+            .expect("the request must reach a terminal failed state rather than time out");
+    match failed_status {
+        RequestStatusDTO::Failed { reason } => {
+            let reason = reason.expect("a failed upgrade must carry a reason");
+            // The station wraps the management-canister error in this prefix
+            // (see ChangeExternalCanisterRequestExecute), so its presence
+            // confirms the failure came from the rejected `install_code` call.
+            // The IC's own wording is surfaced in the panic message rather than
+            // asserted on, to avoid coupling the test to replica error strings.
+            assert!(
+                reason.contains("failed to install external canister"),
+                "upgrade failed for an unexpected reason: {reason}"
+            );
+        }
+        other => panic!("expected the upgrade to fail, got {other:?}"),
+    }
 
     // With `keep`, the same upgrade executes successfully.
     submit_request_approval(
--- a/tests/integration/src/dfx_orbit/install.rs
+++ b/tests/integration/src/dfx_orbit/install.rs
@@ -18,9 +18,10 @@ use dfx_orbit::{
         VerifyCanisterArgs,
     },
 };
+use flate2::read::GzDecoder;
 use sha2::{Digest, Sha256};
 use station_api::{GetRequestInput, RequestApprovalStatusDTO};
-use std::io::Write;
+use std::io::{Read, Write};
 use tempfile::NamedTempFile;
 
 fn hash(data: &[u8]) -> Vec<u8> {
@@ -333,8 +334,15 @@ fn canister_upgrade_with_wasm_memory_persistence_keep() {
 
     // Mark the test module as supporting Enhanced Orthogonal Persistence and
     // install it, so that upgrades require `wasm_memory_persistence = keep`.
+    // The bundled module is gzipped and must be decompressed before appending
+    // the custom section: the IC reads the uncompressed size of gzipped
+    // modules from the trailing bytes of the gzip stream, so bytes appended
+    // after the stream would corrupt the module.
     let test_canister = create_canister(&env, canister_ids.station);
-    let mut module_bytes = get_canister_wasm("test_canister");
+    let mut module_bytes = Vec::new();
+    GzDecoder::new(get_canister_wasm("test_canister").as_slice())
+        .read_to_end(&mut module_bytes)
+        .unwrap();
     append_wasm_custom_section(
         &mut module_bytes,
         "icp:private enhanced-orthogonal-persistence",
--- a/tests/integration/src/dfx_orbit/install.rs
+++ b/tests/integration/src/dfx_orbit/install.rs
@@ -29,6 +29,36 @@ fn hash(data: &[u8]) -> Vec<u8> {
     hasher.finalize().to_vec()
 }
 
+/// Appends a custom section to a Wasm module. The IC decides whether a module
+/// supports Enhanced Orthogonal Persistence by the presence of the
+/// `icp:private enhanced-orthogonal-persistence` custom section, so appending
+/// it turns the plain test canister into an EOP module from the IC's
+/// perspective.
+fn append_wasm_custom_section(module: &mut Vec<u8>, name: &str, payload: &[u8]) {
+    fn write_leb128(mut value: usize, out: &mut Vec<u8>) {
+        loop {
+            let mut byte = (value & 0x7f) as u8;
+            value >>= 7;
+            if value != 0 {
+                byte |= 0x80;
+            }
+            out.push(byte);
+            if value == 0 {
+                break;
+            }
+        }
+    }
+
+    let mut contents = Vec::new();
+    write_leb128(name.len(), &mut contents);
+    contents.extend_from_slice(name.as_bytes());
+    contents.extend_from_slice(payload);
+
+    module.push(0); // custom section id
+    write_leb128(contents.len(), module);
+    module.extend_from_slice(&contents);
+}
+
 /// Test installing a canister through orbit using the station agent
 #[test]
 fn canister_install_no_chunks() {
@@ -279,6 +309,117 @@ fn canister_upgrade_options_round_trip() {
     wait_for_request(&env, other_user, canister_ids.station, request).unwrap();
 
     // The canister still runs the same module after the upgrade.
+    let status = canister_status(&env, Some(canister_ids.station), test_canister);
+    assert_eq!(status.module_hash, Some(module_hash));
+}
+
+/// Test upgrading a canister that runs a module declaring Enhanced Orthogonal
+/// Persistence — the actual use case motivating `--wasm-memory-persistence`.
+/// The IC refuses to upgrade such canisters unless
+/// `wasm_memory_persistence = keep` is set (a safety check against
+/// accidentally dropping their main memory), so the same upgrade request must
+/// fail without the flag and execute successfully with it.
+#[test]
+fn canister_upgrade_with_wasm_memory_persistence_keep() {
+    let TestEnv {
+        mut env,
+        canister_ids,
+        ..
+    } = setup_new_env();
+
+    let (_dfx_user, _) = setup_dfx_user(&env, &canister_ids);
+    let other_user = user_test_id(1);
+    add_user(&env, other_user, vec![], canister_ids.station);
+
+    // Mark the test module as supporting Enhanced Orthogonal Persistence and
+    // install it, so that upgrades require `wasm_memory_persistence = keep`.
+    let test_canister = create_canister(&env, canister_ids.station);
+    let mut module_bytes = get_canister_wasm("test_canister");
+    append_wasm_custom_section(
+        &mut module_bytes,
+        "icp:private enhanced-orthogonal-persistence",
+        &[],
+    );
+    let module_hash = hash(&module_bytes);
+    env.install_canister(
+        test_canister,
+        module_bytes.clone(),
+        vec![],
+        Some(canister_ids.station),
+    );
+
+    permit_change_operation(&env, &canister_ids);
+    set_four_eyes_on_change(&env, &canister_ids);
+
+    let config = DfxOrbitTestConfig {
+        canister_ids: vec![(String::from("test"), test_canister)],
+        ..Default::default()
+    };
+
+    let mut wasm = NamedTempFile::new().unwrap();
+    wasm.write_all(&module_bytes).unwrap();
+
+    let install_args = |wasm_memory_persistence| RequestCanisterInstallArgs {
+        canister: String::from("test"),
+        mode: CanisterInstallModeArgs::Upgrade,
+        wasm_memory_persistence,
+        skip_pre_upgrade: false,
+        wasm: wasm.path().as_os_str().to_str().unwrap().to_string(),
+        argument: None,
+        arg_file: None,
+        asset_canister: None,
+    };
+    let args_without_keep = install_args(None);
+    let args_with_keep = install_args(Some(WasmMemoryPersistenceArgs::Keep));
+
+    let (request_without_keep, request_with_keep) = dfx_orbit_test(&mut env, config, async {
+        let dfx_orbit = setup_dfx_orbit(canister_ids.station).await;
+
+        let mut requests = Vec::new();
+        for args in [args_without_keep, args_with_keep] {
+            let request = RequestArgs {
+                title: None,
+                summary: None,
+                action: RequestArgsActions::Canister(RequestCanisterArgs {
+                    action: RequestCanisterActionArgs::Install(args),
+                }),
+            }
+            .into_request(&dfx_orbit)
+            .await
+            .unwrap();
+
+            let response = dfx_orbit.station.request(request).await.unwrap();
+            requests.push(response.request);
+        }
+
+        let with_keep = requests.pop().unwrap();
+        let without_keep = requests.pop().unwrap();
+        (without_keep, with_keep)
+    });
+
+    // Without `keep`, the IC refuses to drop the main memory of a canister
+    // running an Enhanced Orthogonal Persistence module, so the upgrade fails.
+    submit_request_approval(
+        &env,
+        other_user,
+        canister_ids.station,
+        request_without_keep.clone(),
+        RequestApprovalStatusDTO::Approved,
+    );
+    wait_for_request(&env, other_user, canister_ids.station, request_without_keep)
+        .expect_err("the IC must reject an EOP upgrade without wasm_memory_persistence = keep");
+
+    // With `keep`, the same upgrade executes successfully.
+    submit_request_approval(
+        &env,
+        other_user,
+        canister_ids.station,
+        request_with_keep.clone(),
+        RequestApprovalStatusDTO::Approved,
+    );
+    wait_for_request(&env, other_user, canister_ids.station, request_with_keep).unwrap();
+
+    // The canister still runs the (EOP-marked) module after the upgrade.
     let status = canister_status(&env, Some(canister_ids.station), test_canister);
     assert_eq!(status.module_hash, Some(module_hash));
 }
--- a/tests/integration/src/dfx_orbit/install.rs
+++ b/tests/integration/src/dfx_orbit/install.rs
@@ -339,22 +339,30 @@ fn canister_upgrade_with_wasm_memory_persistence_keep() {
     // modules from the trailing bytes of the gzip stream, so bytes appended
     // after the stream would corrupt the module.
     let test_canister = create_canister(&env, canister_ids.station);
-    let mut module_bytes = Vec::new();
+    let mut installed_module_bytes = Vec::new();
     GzDecoder::new(get_canister_wasm("test_canister").as_slice())
-        .read_to_end(&mut module_bytes)
+        .read_to_end(&mut installed_module_bytes)
         .unwrap();
     append_wasm_custom_section(
-        &mut module_bytes,
+        &mut installed_module_bytes,
         "icp:private enhanced-orthogonal-persistence",
         &[],
     );
-    let module_hash = hash(&module_bytes);
+    let installed_module_hash = hash(&installed_module_bytes);
     env.install_canister(
         test_canister,
-        module_bytes.clone(),
+        installed_module_bytes.clone(),
         vec![],
         Some(canister_ids.station),
     );
+
+    // Upgrade to a *different* module — still EOP-marked, but carrying an extra
+    // marker section so its hash differs from the installed one. That lets the
+    // post-upgrade module hash distinguish "new code installed" from a no-op.
+    let mut upgraded_module_bytes = installed_module_bytes.clone();
+    append_wasm_custom_section(&mut upgraded_module_bytes, "orbit:upgrade-marker", &[1]);
+    let upgraded_module_hash = hash(&upgraded_module_bytes);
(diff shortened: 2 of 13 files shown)


tools/dfx-orbit/src/canister/install.rs
```diff
@@ -160,6 +177,45 @@ impl RequestCanisterInstallArgs {
...
+            CanisterInstallModeArgs::Reinstall => {
+                self.ensure_no_upgrade_options()?;
+                Ok(CanisterInstallMode::Reinstall)
+            }
+            CanisterInstallModeArgs::Upgrade => {
+                // Collapse to `None` when neither flag is set so the request stays
+                // byte-for-byte identical to a plain `--mode upgrade`.
+                let options = if self.wasm_memory_persistence.is_some() || self.skip_pre_upgrade {
+                    Some(CanisterUpgradeOptionsInput {
+                        wasm_memory_persistence: self.wasm_memory_persistence.map(Into::into),
```
daniel-wong-dfinity-org-twin: IMHO, Into is evil. From is good.

The funny thing is, Rust agrees that naming the destination type is good when it comes to converting integers: you cannot do `some_u64 + some_u32`, even though "obviously" what this needs is to simply upgrade the `u32` to a `u64`, and this is in fact what C does (it does some conversions that are questionable though).

> Explicit is better than implicit.
> --Zen of Python

The problem with implicit is that you are assuming that necessary information is already understood. Well, as they say, assuming is how you make an ass out of u and me. Such assumptions are convenient to the writer, but we should optimize for the reader, at least, in the opinion of the Google style guides.

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -160,6 +177,45 @@ impl RequestCanisterInstallArgs {
...
+                        skip_pre_upgrade: self.skip_pre_upgrade.then_some(true),
```
daniel-wong-dfinity-org-twin: IMHO .then_some is pretty esoteric (e.g. in the ic repo, it is used "only" 113 times), and not very self-explanatory. Furthermore, every Rust programmer knows

```
if self.skip_pre_upgrade {
    Some(true)
} else {
    None
}
```

and it is not THAT verbose. Therefore `.then_some` is doing only a very small amount of "lifting".

Since `.then_some` is esoteric and only does I small amount of lifting, I say that it has a low "power to weight" ratio. Therefore, I reckon that the "primitive" way (i.e. composing `if`, `Some`, `None`, etc.) is best.

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -160,6 +177,45 @@ impl RequestCanisterInstallArgs {
...
+                    })
+                } else {
+                    None
+                };
+                Ok(CanisterInstallMode::Upgrade(options))
+            }
+        }
+    }
+
+    fn ensure_no_upgrade_options(&self) -> anyhow::Result<()> {
```
daniel-wong-dfinity-org-twin: At least to me, "ensure" does not mean "freak out if there is a bad thing". I think what it means "do whatever you need to do so that there will be no bad thing". E.g. if I want to super ensure the eggs survive the journey home from the grocery store, I wrap them in several layers of bubble wrap. "ensure" does not mean that I chuck them into the back seat of the car, and scream if I find that some of them are cracked when I get home.

OTOH, after looking at a bunch of examples in the ic repo, I see that in an overwhelming majority of cases, when a function name begins with "ensure_", it returns Result.

In any case, here is my suggestion (which you might consider bike shedding):


```suggestion
    fn err_if_upgrade_flags_were_passed(&self) -> anyhow::Result<()> {
```

To understand what this does, the reader need not know ANY conventions (such as ensure = Result, which isn't even 100% followed). This makes it difficult to misunderstand. "Difficult to misunderstand" is a level of clarity that can hardly be exceeded!

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -193,6 +249,26 @@ impl From<CanisterInstallMode> for CanisterInstallModeArgs {
     }
 }
 
+/// Whether the canister's Wasm main memory is kept or replaced on upgrade.
+/// Equivalent to `dfx canister install --wasm-memory-persistence XXX` and
+/// `orbit_station_api::WasmMemoryPersistence`.
+#[derive(Copy, Clone, Eq, PartialEq, Debug, ValueEnum)]
+pub enum WasmMemoryPersistenceArgs {
```
daniel-wong-dfinity-org-twin, line 263: I guess the local pattern is to duplicate enums (and should be continued). Nevertheless, I have to say this (to the whole pattern, not just this instance of it): "eww".

The conversion from Keep to Keep and Replace to Replace has an extremely high "you don't say!" factor, and therefore, pretty clearly shows that this is pathological.

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -213,10 +289,128 @@ impl DfxOrbit {
         };
         writeln!(output, "Mode: {mode}")?;
 
+        if let CanisterInstallMode::Upgrade(Some(options)) = &op.mode {
+            if let Some(persistence) = &options.wasm_memory_persistence {
+                let persistence = match persistence {
+                    WasmMemoryPersistence::Keep => "keep",
+                    WasmMemoryPersistence::Replace => "replace",
+                };
```
daniel-wong-dfinity-org-twin: I know that this just uses the same pattern as existing code (so just keep on doing this), but it was not a good pattern to begin with. What we are doing here is converting to a string. Such functionality is so fundamental that has optimized it to be LESS THAN 1 line: `#[derive(Debug)]`. Let's use that, as god intended:

```
writeln!(output, "Wasm memory persistence: {persistence:?}");
```

The only "problem" with this is that instead of "keep", you get "Keep". If you really hate the capital "K", you can get rid of it very concisely as follows:

```
let persistence = format!("{:?}", persistence).to_lowercase();
writeln!(output, "Wasm memory persistence: {persistence}");
```

tools/dfx-orbit/src/canister.rs
```diff
@@ -9,7 +9,8 @@ mod util;
 
 pub use self::{
```
daniel-wong-dfinity-org-twin: I didn't even know this was possible. Seems unnecessary. I.e. instead of

```
use self::{a::A1, a::A2, b::B};
```

the way I've always seen it done is

```
use a::{A1, A2};
use b::B;
```

tools/dfx-orbit/src/canister.rs
daniel-wong-dfinity-org-twin, line 1: I am a  bit confused how it is possible that the only changes in a .rs file is just more `use`. It seems to me, the newly `use`ed things aren't even being.... USED! Doesn't the compiler warn about unused `use`s??

tools/dfx-orbit/README.md
```diff
@@ -141,6 +141,31 @@ Then a verifier can verify this request, using:
 dfx-orbit verify [REQUEST_ID] canister install --mode upgrade [CANISTER_NAME] --wasm [WASM_PATH]
 ```
 
+##### Preserving main memory across upgrades
+
+When upgrading, you can control what happens to the canister's Wasm main memory
+with `--wasm-memory-persistence`:
+
+- `keep` — preserve the main memory. This is **required** for Motoko canisters
+  that use [Enhanced Orthogonal Persistence](https://internetcomputer.org/docs/current/motoko/main/canister-maintenance/orthogonal-persistence/enhanced);
```
daniel-wong-dfinity-org-twin: This URL doesn't work for me.

tools/dfx-orbit/README.md
```diff
@@ -141,6 +141,31 @@ Then a verifier can verify this request, using:
 dfx-orbit verify [REQUEST_ID] canister install --mode upgrade [CANISTER_NAME] --wasm [WASM_PATH]
 ```
 
+##### Preserving main memory across upgrades
+
+When upgrading, you can control what happens to the canister's Wasm main memory
+with `--wasm-memory-persistence`:
```
daniel-wong-dfinity-org-twin, line 147: Ideally, dfx-orbit should match dfx. OTOH, dfx is deprecated...

I'm confused how dfx behaves on this. On the one hand, it says [here][infer] that dfx infers the correct value when it detects that the wasm needs wasm memory persistence. I like that, because otherwise, it is way too easy to forget, and the consequences are irrecoverable.

OTOH, [here][wasm-flag] it says that some checks are skipped when using `--wasm`. Therefore, I'm not sure how `--wasm` affects wasm memory persistence inference (in dfx)...

[infer]: https://github.com/dfinity/sdk/blob/master/CHANGELOG.md#feat-support-for-motokos-enhanced-orthogonal-persistence

[wasm-flag]: https://github.com/dfinity/sdk/blob/master/CHANGELOG.md#feat-install-arbitrary-wasm-into-canisters

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -213,10 +289,128 @@ impl DfxOrbit {
...
+                .unwrap(),
+            CanisterInstallMode::Install
+        );
+        assert_eq!(
+            args(CanisterInstallModeArgs::Reinstall)
+                .install_mode()
+                .unwrap(),
+            CanisterInstallMode::Reinstall
+        );
+    }
```
daniel-wong-dfinity-org-twin: To beat a dead horse, look at this test. All it is telling us is that Install maps to Install (and Reinstall maps to Reinstall). Sure, AI can effortlessly spew out this type of code. I don't think that suddenly makes highly redundant code "ok".

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -213,10 +289,128 @@ impl DfxOrbit {
...
         writeln!(output, "Module checksum: {}", &op.module_checksum)?;
         if let Some(arg_checksum) = &op.arg_checksum {
             writeln!(output, "Argument checksum: {arg_checksum}")?;
         }
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
```
daniel-wong-dfinity-org-twin, line 318: Is inlining mod tests the usual way around here? If not, spinning out is helpful to build systems when only tests change.

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -213,10 +289,128 @@ impl DfxOrbit {
...
+
+    #[test]
+    fn upgrade_flags_rejected_for_install_and_reinstall() {
+        for mode in [
+            CanisterInstallModeArgs::Install,
+            CanisterInstallModeArgs::Reinstall,
+        ] {
+            let mut with_persistence = args(mode);
+            with_persistence.wasm_memory_persistence = Some(WasmMemoryPersistenceArgs::Keep);
+            assert!(with_persistence.install_mode().is_err());
```
daniel-wong-dfinity-org-twin: ok, but, what is the Err saying? Is it really occurring because of an incompatibility?

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -213,10 +289,128 @@ impl DfxOrbit {
...
+        );
+    }
+
+    #[test]
+    fn upgrade_flags_rejected_for_install_and_reinstall() {
+        for mode in [
+            CanisterInstallModeArgs::Install,
+            CanisterInstallModeArgs::Reinstall,
+        ] {
+            let mut with_persistence = args(mode);
```
daniel-wong-dfinity-org-twin: variable names should be noun-y. E.g.


```suggestion
            let mut install_args_with_wasm_memory_persistence = args(mode);
```

tools/dfx-orbit/src/canister/install.rs
```diff
@@ -213,10 +289,128 @@ impl DfxOrbit {
...
+            with_persistence.wasm_memory_persistence = Some(WasmMemoryPersistenceArgs::Keep);
+            assert!(with_persistence.install_mode().is_err());
+
+            let mut with_skip = args(mode);
```
daniel-wong-dfinity-org-twin: Ditto.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+        canister_ids: vec![(String::from("test"), test_canister)],
+        ..Default::default()
+    };
+
+    let mut wasm = NamedTempFile::new().unwrap();
+    let module_bytes = get_canister_wasm("test_canister");
+    wasm.write_all(&module_bytes).unwrap();
+
+    let inner_args = RequestCanisterInstallArgs {
+        canister: String::from("test"),
```
daniel-wong-dfinity-org-twin, line 234: This is not the way I'm used to seeing it. I mean, it is perfectly readable otherwise, but if this is not the usual way around here (idk), do it the usual way.

Remind me to stab Rust for not having a String literal.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+        mode: CanisterInstallModeArgs::Upgrade,
+        wasm_memory_persistence: Some(WasmMemoryPersistenceArgs::Keep),
+        skip_pre_upgrade: true,
+        wasm: wasm.path().as_os_str().to_str().unwrap().to_string(),
```
daniel-wong-dfinity-org-twin, line 238: Again, Rust really needs to be stabbed in the face.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+    let config = DfxOrbitTestConfig {
+        canister_ids: vec![(String::from("test"), test_canister)],
+        ..Default::default()
+    };
+
+    let mut wasm = NamedTempFile::new().unwrap();
+    let module_bytes = get_canister_wasm("test_canister");
+    wasm.write_all(&module_bytes).unwrap();
+
+    let inner_args = RequestCanisterInstallArgs {
```
daniel-wong-dfinity-org-twin:
```suggestion
    let install_args = RequestCanisterInstallArgs {
```

The name of the type is of the form *Args -> it makes no sense for the variable to be called ***inner***_args.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+
+    dfx_orbit_test(&mut env, config, async {
+        let dfx_orbit = setup_dfx_orbit(canister_ids.station).await;
+
+        let request = RequestArgs {
+            title: None,
+            summary: None,
+            action: RequestArgsActions::Canister(RequestCanisterArgs {
+                action: RequestCanisterActionArgs::Install(inner_args.clone()),
+            }),
```
daniel-wong-dfinity-org-twin, line 252: Not in this PR, but in principle...

Rant: Wrap [ALL the things][all-the-things]!

[all-the-things]: https://media.npr.org/assets/img/2013/11/12/clean-all-the-things_wide-1ba922355d68293ef047d8c7e68289a6555e19b3.jpg?s=1400

Seriously, `From` should be implemented, or something, because then, this abomination simply becomes

```suggestion
            action: RequestArgsAction::from(args),
```

Nobody wants or needs to see how the sausage is made. They just want to know, "you give me meat bits, and I'll give you a completed sausage, ready for grilling".

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+        }
+        .into_request(&dfx_orbit)
+        .await
```
daniel-wong-dfinity-org-twin, line 255: Not in this PR (since this predates this PR, and is therefore out of scope), but in principle...

No sane human would ever expect a conversion to be async.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+        .unwrap();
+
+        let request = dfx_orbit.station.request(request.clone()).await.unwrap();
```
daniel-wong-dfinity-org-twin: Um, if you send a request (which is what it looks like here, because `request` is being passed, and then `.await`ed) the result should be a ***response***. Therefore, it is super baffling that the result is being shoe horned into a variable named `request`.

I mean, if this is how it's being done around here, keep on doing it, but just be aware that this is forcing everyone to hold their noses as they read this...

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+
+        let req_response = dfx_orbit
```
daniel-wong-dfinity-org-twin: Everyone knows that a response comes as the result of a request. That goes without saying. Therefore, it is redundant to have a variable named ***req***_response. That prefix adds no useful information.

It is certainly a good idea to have a more specific name than `response`. The usual (and good) way to pick such names is to name it after the OPERATION. In this case, that's `review`. Thus, this would be named `review_response`.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -152,3 +156,107 @@ fn canister_install(use_chunks: bool) {
...
+        .verify(&dfx_orbit, &req_response)
+        .await
+        .expect_err("verification must reject a mismatched wasm_memory_persistence");
+
+        request.request
+    });
+
+    // The upgrade request is still pending, so the canister remains empty.
+    let status = canister_status(&env, Some(canister_ids.station), test_canister);
+    assert_eq!(status.module_hash, None);
```
daniel-wong-dfinity-org-twin: Wouldn't it make more sense to follow through, i.e. install the wasm? I mean, ultimately, what we really want to see is that `Keep` and `skip_pre_upgrade` were actually, you know, PERFORMED as requested, right??

apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
```diff
@@ -0,0 +1,78 @@
...
+  set: value => {
+    switch (value) {
+      case 'keep':
+        emit('update:modelValue', { keep: null });
+        break;
+      case 'replace':
+        emit('update:modelValue', { replace: null });
+        break;
+      default:
+        emit('update:modelValue', undefined);
```
daniel-wong-dfinity-org-twin, line 70: This looks like sweeping a problem under the rug.

daniel-wong-dfinity-org-twin (commented):
For the record, I really can't properly review the TS stuff. Nevertheless, I muddled through that portion of the review as best as I can, since reviewers are scarce these days...

tools/dfx-orbit/src/canister.rs
aterga: They are `pub use` re-exports, not plain `use` — they make `WasmMemoryPersistenceArgs` part of the crate's public API (`dfx_orbit::canister::WasmMemoryPersistenceArgs`), which the integration tests in `tests/integration/src/dfx_orbit/install.rs` consume. The compiler only warns about unused *private* imports; re-exports are themselves the "use".

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/README.md
aterga: dfx infers it from the Wasm itself: Motoko EOP modules carry an `enhanced-orthogonal-persistence` metadata section, and when dfx sees it, it sets `wasm_memory_persistence = keep` automatically (that's why the `--wasm` fast path, which skips module inspection, can behave differently). dfx-orbit keeps the flag explicit for now because requests are approved by third parties — the verifier checks the request against flags the operator states, rather than against something inferred client-side. Inferring a default from the module metadata (and warning when it disagrees with the flag) would be a nice follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Yes — inline `#[cfg(test)] mod tests` is the pattern throughout this repo (e.g. `core/station/impl/src/mappers/request_operation.rs`); there are no separate unit-test files in `dfx-orbit`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Done in b88cd2e — the test now installs the module first, then the approved upgrade executes with `wasm_memory_persistence = replace` + `skip_pre_upgrade` and `wait_for_request` asserts it completed, so the options are exercised through the management canister rather than left pending. One caveat: it uses `replace` instead of `keep`, because the test canister isn't built with Enhanced Orthogonal Persistence and the IC rejects `keep` for modules without the EOP metadata section. Truly proving `keep` preserves memory would need a Motoko EOP fixture wasm, which felt out of scope here.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Done in b88cd2e — now `.map(WasmMemoryPersistence::from)`, naming the destination type.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Done in b88cd2e — replaced `.then_some(true)` with the explicit `if self.skip_pre_upgrade { Some(true) } else { None }`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Applied your suggestion verbatim in b88cd2e — the function is now `err_if_upgrade_flags_were_passed`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Agreed the identity arms read as "you don't say!", but the duplication is load-bearing here: the CLI-side enum carries clap's `ValueEnum` derive, which the transport crate (`station_api`) shouldn't depend on. Kept the local pattern, per your note.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Done in b88cd2e — the display string is now derived via `format!("{persistence:?}").to_lowercase()`, as suggested.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister.rs
aterga: Done in b88cd2e — split into per-module `pub use` statements.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/README.md
aterga: Fixed in b88cd2e — the IC docs moved wholesale to docs.internetcomputer.org, which broke the old path. New link (verified it resolves): https://docs.internetcomputer.org/motoko/orthogonal-persistence/enhanced

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Fair — folded it into the single `plain_modes_produce_no_upgrade_options` test in b88cd2e. What it actually guards is the flag-validation path (install/reinstall *without* flags must return `Ok` rather than trip `err_if_upgrade_flags_were_passed`); the identity mapping is just how that's observable.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Done in b88cd2e — the test now asserts the error message is the "--wasm-memory-persistence and --skip-pre-upgrade are only valid with --mode upgrade" rejection, not just any `Err`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Applied verbatim in b88cd2e — and its sibling below is now `install_args_with_skip_pre_upgrade`, covering the "Ditto".

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: It matches the sibling `canister_install` test in this same file (which uses `String::from("test")`), so I kept it for local consistency — same goes for the `.to_str()` dance on the temp-file path below.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Applied in b88cd2e — renamed to `install_args`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Agreed — a `From` impl (or a small builder) on the args wrappers would clean up all these call sites. Leaving it as a follow-up since the wrapping pattern predates this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Renamed in b88cd2e — the submit result now lands in `response`, and the closure returns `response.request`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Renamed to `review_response` in b88cd2e, following the operation-based naming you suggested.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
aterga: Done in b88cd2e — the catch-all `default:` arm is gone; `'default'` is now an explicit case with a comment explaining it means "let the IC decide" and maps to an absent `wasm_memory_persistence` option in the request. The select's items are constrained to the three options, so every case is now spelled out.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Agreed on the principle — the twist here is that `into_request` isn't actually a pure conversion: for `canister install --asset-canister` it uploads the wasm chunks to the asset canister (`ic_asset::upload`), and the settings path does canister-ID lookups, which is why it's `async`. So the real sin is the name underselling the I/O — something like `build_request`/`prepare_request` would be more honest. Since it predates this PR, leaving that rename for a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Applied in b88cd2e — this one is now `install_args_with_skip_pre_upgrade`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: No argument here — the `NamedTempFile` → `&str` → `String` dance is pure ceremony. Kept it as-is since it mirrors the sibling `canister_install` test in this file; a shared helper could clean up both in a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
daniel-wong-dfinity-org-twin: ~LG. I cannot find the "Mark resolved" button anymore... GitHub bug??~

Update:

Ah. I found it!

However, when I click it, I get "Failed to resolve thread"! Can GitHub do anything correctly? Or are they just vibing everything now???

Here's what I'll do in lieu of "Resolve comment": I'll thumbs up your reply if I consider the thread as resolved...

tools/dfx-orbit/src/canister/install.rs
daniel-wong-dfinity-org-twin: > load-bearing

FYI: This sounds very AI.

Yes, I know that this is here for clap. That does NOT mean we should quietly surrender to copy & paste.

tools/dfx-orbit/src/canister/install.rs
daniel-wong-dfinity-org-twin: Ok, but how about mode? That's why I said,

> I know that this just uses the same pattern as existing code (so ***just keep on doing this***)

This thread was more of a rant than an actual suggestion.

tools/dfx-orbit/src/canister/install.rs
daniel-wong-dfinity-org-twin: This was not asking for any change. This was just pointing out how silly it is to have both CanisterInstallModeArgs and CanisterInstallMode.

Whether its one test or two is orthogonal to the point. I slightly prefer two, but it's a bit of rearranging deck chairs on the Titanic.

tools/dfx-orbit/src/canister.rs
daniel-wong-dfinity-org-twin: Ah. I missed the `pub`.

tools/dfx-orbit/README.md
daniel-wong-dfinity-org-twin: Just to be clear, this thread is not asking for any changes in this PR (hence 👍). Just discussing.

> approved by third parties

This fact does not significantly reduce the danger of footgun, which is my main concern in this thread.

I'm fine with this PR not having detection, because at least this makes it POSSIBLE to have wasm memory persistence. I'm more saying, ideally, if we had infinite resources, we would improve this in a later PR, because the promise of dfx-orbit has always been "dfx-orbit is 'just like' dfx".

> warning

Yes, this would be very good.

Not sure what dfx does, but what I would prefer is if dfx actually BLOCKS things that look very much like mistakes, not just warn, but with a bypass (i.e. `--force`). But as I said, it is deprecated, which means, if it doesn't already do that, then it probably never will...

tools/dfx-orbit/README.md
daniel-wong-dfinity-org-twin: Do as you see fit (including nothing).

Might be even better to link to the most relevant section, "IC main memory retention", but this is not critical by any means, and you could argue that section links are too unstable for this to really "make sense".

tests/integration/src/dfx_orbit/install.rs
daniel-wong-dfinity-org-twin: > out of scope here.

I don't think so. The whole motivation here is so that I can install a canister that was recently written in Motoko (and therefore, uses EOP).

apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
daniel-wong-dfinity-org-twin: Does the IC decide automatically?? If so, then, it seems that this feature is not super necessary.

tests/integration/src/dfx_orbit/install.rs
aterga: Fair point — done in 595ddfd. New integration test `canister_upgrade_with_wasm_memory_persistence_keep`: it marks the test module as EOP by appending the `icp:private enhanced-orthogonal-persistence` custom section (which is exactly how the IC detects EOP support — no Motoko toolchain needed), installs it, and then shows the same upgrade request **fails without `keep`** (the IC's safety check refuses to drop an EOP canister's main memory) and **executes successfully with `keep`**. That exercises the real motivating scenario end-to-end through the station.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/wallet/src/components/inputs/CanisterWasmMemoryPersistenceSelect.vue
aterga: No — the IC never infers `keep`. What it does is fail closed: upgrading a canister whose module declares EOP *without* `keep` is rejected outright (so no silent memory loss, but also no upgrade). It's dfx that infers client-side from the Wasm metadata and fills the flag in for you. So the option is necessary — without it, EOP canisters simply cannot be upgraded through Orbit at all. The new integration test in 595ddfd demonstrates both halves (rejected without `keep`, succeeds with it).

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Noted — the resolve API does work from my side, so I've resolved this thread and the `pub use` one you closed ("Ah. I missed the pub"). If you 👍 any of the other replies, I can resolve those threads too.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: The constraint is the orphan rule: `dfx-orbit` can't `impl clap::ValueEnum for station_api::WasmMemoryPersistence` (foreign trait on a foreign type). The non-duplicating options are (a) a feature-gated `clap` dependency in `station_api` so the API enum derives `ValueEnum` directly, or (b) a newtype wrapper in the CLI with a hand-written `ValueEnum` impl — which ends up as roughly the same amount of ceremony. Happy to do (a) as a follow-up; it would eliminate this pattern for all the `*Args` enums here, not just this one.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Understood. For `mode` specifically, the Debug trick doesn't quite work: the `Upgrade` variant carries a payload, so `{mode:?}` would print `Upgrade(Some(CanisterUpgradeOptionsInput { .. }))` rather than a clean `Upgrade` — hence the hand-written match stays there.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Understood — leaving the test as-is then. The `CanisterInstallModeArgs`/`CanisterInstallMode` pairing itself is the orphan-rule story from the other thread; a feature-gated `clap` in `station_api` would dissolve both enums into one as a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/README.md
aterga: Done in 595ddfd — the link now points at the "[IC main memory retention](https://docs.internetcomputer.org/motoko/orthogonal-persistence/enhanced#ic-main-memory-retention)" section, which is exactly the part describing `wasm_memory_persistence`. If the anchor ever rots, it degrades gracefully to the page top.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/README.md
aterga: Agreed on the follow-up direction. One mitigating fact: the footgun is smaller than it looks because the IC itself fails closed — upgrading a canister that runs an EOP module without `keep` is rejected outright, so the worst case is a blocked request, not lost memory (the new integration test in 595ddfd pins this down). The residual gap is the classical→EOP first upgrade, where nothing forces the flag. A dfx-orbit check that reads the Wasm metadata and blocks with a `--force` escape hatch (your preference, which I share — the request is immutable once submitted, so warn-and-proceed is too weak here) would close that; noting it as the follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: > This sounds very AI.

> _Generated by [Claude Code](https://claude.ai/code)_

indeed.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -29,6 +30,36 @@ fn hash(data: &[u8]) -> Vec<u8> {
     hasher.finalize().to_vec()
 }
 
+/// Appends a custom section to a Wasm module. The IC decides whether a module
+/// supports Enhanced Orthogonal Persistence by the presence of the
+/// `icp:private enhanced-orthogonal-persistence` custom section, so appending
+/// it turns the plain test canister into an EOP module from the IC's
+/// perspective.
+fn append_wasm_custom_section(module: &mut Vec<u8>, name: &str, payload: &[u8]) {
```
daniel-wong-dfinity-org-twin, line 38: FYI

I do not know wasm in sufficient detail to say whether this is correct, but fwiw it seems very plausible.

daniel-wong-dfinity-org-twin, line 38: Do as you see fit.

There is probably some library that does this. I guess it's ok to roll our own here, because this is "just test".

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -282,3 +313,121 @@ fn canister_upgrade_options_round_trip() {
     let status = canister_status(&env, Some(canister_ids.station), test_canister);
     assert_eq!(status.module_hash, Some(module_hash));
 }
+
+/// Test upgrading a canister that runs a module declaring Enhanced Orthogonal
+/// Persistence — the actual use case motivating `--wasm-memory-persistence`.
+/// The IC refuses to upgrade such canisters unless
+/// `wasm_memory_persistence = keep` is set (a safety check against
+/// accidentally dropping their main memory), so the same upgrade request must
+/// fail without the flag and execute successfully with it.
```
daniel-wong-dfinity-org-twin, line 322: Thanks. This is nice. I guess Claude did this. If so, kudos to AI  🥂

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -282,3 +313,121 @@ fn canister_upgrade_options_round_trip() {
...
+            .into_request(&dfx_orbit)
+            .await
+            .unwrap();
+
+            let response = dfx_orbit.station.request(request).await.unwrap();
+            requests.push(response.request);
+        }
+
+        let with_keep = requests.pop().unwrap();
+        let without_keep = requests.pop().unwrap();
```
daniel-wong-dfinity-org-twin: Do as you see fit.

assert requests empty?

this kind of shows why it is a bit weird to have this 2 element vec. I think the way to avoid this is to turn the loop into a lambda. Then, here,

```
let response_with_keep = install_code(args_with_keep);
let response_without_keep = install_code(args_without_keep);
```

Then, there is no need to assert empty.

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -282,3 +313,121 @@ fn canister_upgrade_options_round_trip() {
...
+    // running an Enhanced Orthogonal Persistence module, so the upgrade fails.
+    submit_request_approval(
+        &env,
+        other_user,
+        canister_ids.station,
+        request_without_keep.clone(),
+        RequestApprovalStatusDTO::Approved,
+    );
+    wait_for_request(&env, other_user, canister_ids.station, request_without_keep)
+        .expect_err("the IC must reject an EOP upgrade without wasm_memory_persistence = keep");
```
daniel-wong-dfinity-org-twin: Can we inspect the err more closely to make sure that it failed for the right reason?

tests/integration/src/dfx_orbit/install.rs
```diff
@@ -282,3 +313,121 @@ fn canister_upgrade_options_round_trip() {
...
+    );
+
+    permit_change_operation(&env, &canister_ids);
+    set_four_eyes_on_change(&env, &canister_ids);
+
+    let config = DfxOrbitTestConfig {
+        canister_ids: vec![(String::from("test"), test_canister)],
+        ..Default::default()
+    };
+
```
daniel-wong-dfinity-org-twin, line 374: Can we modify module_bytes again so that after the upgrade we can tell whether new code was installed?

tools/dfx-orbit/src/canister/install.rs
daniel-wong-dfinity-org-twin: No, don't bother. The cure is worse than the disease. Nevertheless, there is a disease here. That's my only point. Hence, "FYI".

daniel-wong-dfinity-org-twin (approved):

tests/integration/src/dfx_orbit/install.rs
aterga: Done in a4d72ce — replaced the two-element `Vec`/`pop` with a `submit_upgrade` closure, so each request is bound directly:

```rust
let request_without_keep = submit_upgrade(args_without_keep).await;
let request_with_keep = submit_upgrade(args_with_keep).await;
```

No more `requests.pop().unwrap()` and no empty-assert needed.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Done in a4d72ce — it now matches the terminal `RequestStatusDTO::Failed { reason }` and asserts the reason contains the station's `"failed to install external canister"` execution-error prefix, which confirms the failure came from the rejected `install_code` call (rather than the request being rejected/cancelled at an earlier stage or timing out). I deliberately don't assert on the IC's own error wording — that would couple the test to replica error strings — but the full reason is surfaced in the panic message if it's ever needed.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: Confirmed correct — I validated the encoding out-of-band by round-tripping a module built with this helper through `WebAssembly.validate` (both a short section name and one crossing the 127-byte LEB128 boundary), and the CI run for a4d72ce exercises it for real: the IC accepts the module and applies the EOP safety check, which is only reachable if the custom section is well-formed.

---
_Generated by [Claude Code](https://claude.ai/code)_

aterga: Agreed — `wasm-encoder` would give a typed API for this, but pulling it in just to append one custom section in a test felt heavier than the ~15-line helper. Keeping the hand-rolled version since it's test-only, as you say.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/src/dfx_orbit/install.rs
aterga: 🥂 — it took a CI round to get there, though: the first cut appended the section to the *gzipped* module, and the IC read the tail of the section name as a 1.7 GB module size (fixed in 3cbcdd5 by decompressing first).

---
_Generated by [Claude Code](https://claude.ai/code)_

tools/dfx-orbit/src/canister/install.rs
aterga: Understood — leaving it as-is. Noting the `station_api` clap-feature cleanup as a possible follow-up, not touching it in this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

sea-snake (approved):

tests/integration/src/dfx_orbit/install.rs
aterga: Good catch — done in 25a7e3a. The upgrade now targets a *distinct* module: I take the installed EOP module and append an extra `orbit:upgrade-marker` custom section (still EOP-marked, so `keep` is still required) whose hash differs from the installed one (`assert_ne!` guards that). The test then asserts both halves:

- after the keep-less upgrade fails, the canister still has `installed_module_hash` (the failed upgrade changed nothing);
- after the `keep` upgrade succeeds, the canister has `upgraded_module_hash` — proving new code was actually installed rather than a no-op.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #656, opened 2026-09-01
ci(release): add Deploy frontend workflow

MRmarioruci:
Phase 2 of 4. Stacked on #655.

Adds the `Deploy frontend` workflow_dispatch and the `scripts/deploy-app` helper it wraps.

You tick which frontends to deploy (wallet, marketing, docs) and run it. The playground job builds each app for playground and uploads it with icx-asset so you can test at the playground URL. `deploy-app` resolves the canister, gets the artifact, unpacks it, syncs the assets with `--no-delete`, and reads the live version meta tag back to confirm what landed. Frontend config is baked in at build time, so playground gets its own build while production ships the released tagged tarball. The signing key comes either from a `DEPLOY_IDENTITY_PEM` environment secret or, locally, straight from 1Password via `--op`, in both cases through a temp pem wiped on exit.

The `production` job is off by default and has no key. Production frontends are deployed with `scripts/deploy-app --op` from a machine that can read the key, which keeps that credential out of this repo entirely. A frontend sync is live to every user the moment it finishes, so we would rather it not be reachable by anything that lands in `.github/`. The job is kept so the wiring exists if that is revisited; with no key it fails fast with a pointer.

`deploy-app` now also verifies the tarball against the `.sha256` published beside it in the release. Releases have always published one and the script was not looking at it, which matters more now that this is a hand-run path with no CI attestation behind it.

The jobs install only what `deploy-app` needs (gh, jq, docker, icx-asset, shasum, curl), so the npm dependency tree and its lifecycle scripts never enter a job that writes a signing key to disk.

Requires the `playground` GitHub Environment with its key, and its deployment branches restricted to `main`. Base of this PR is #655; review that first.

### DIFF
--- a/.github/workflows/release-deploy-frontend.yaml
+++ b/.github/workflows/release-deploy-frontend.yaml
@@ -8,6 +8,10 @@ name: 'Deploy frontend'
 # environment's approval. Backends (station, upgrader, control-panel) are a
 # separate follow-up.
 #
+# These jobs deliberately install as little as possible. deploy-app needs only
+# gh, jq, docker, icx-asset and curl, so there is no Node/pnpm install here and
+# the npm dependency tree never enters the job that handles the signing key.
+#
 # The signing key is read from the `DEPLOY_IDENTITY_PEM` secret. Scope it per
 # environment: put the playground key on the `playground` environment and the
 # production key on the `production` environment, so each job only ever sees its
@@ -57,10 +61,6 @@ jobs:
           # This job downloads via gh, never pushes, so it does not need the
           # GITHUB_TOKEN left in .git/config.
           persist-credentials: false
-      - name: 'Setup Node'
-        uses: ./.github/actions/setup-node
-      - name: 'Install dependencies'
-        run: pnpm install --frozen-lockfile
       - name: 'Set up Docker'
         uses: docker/setup-docker-action@e43656e248c0bd0647d3f5c195d116aacf6fcaf4 # v4
       - name: 'Install icx-asset'
@@ -106,10 +106,6 @@ jobs:
           # This job downloads via gh, never pushes, so it does not need the
           # GITHUB_TOKEN left in .git/config.
           persist-credentials: false
-      - name: 'Setup Node'
-        uses: ./.github/actions/setup-node
-      - name: 'Install dependencies'
-        run: pnpm install --frozen-lockfile
       - name: 'Install icx-asset'
         run: cargo install --locked icx-asset --version 0.29.0
       - name: 'Deploy to production'
--- a/.github/workflows/release-deploy-frontend.yaml
+++ b/.github/workflows/release-deploy-frontend.yaml
@@ -12,8 +12,10 @@ name: 'Deploy frontend'
 # gh, jq, docker, icx-asset, shasum and curl, so there is no Node/pnpm install
 # here and the npm dependency tree never enters the job that handles the key.
 #
-# The signing key is read from the `DEPLOY_IDENTITY_PEM` secret, scoped per
-# environment, so a job only ever sees its own key.
+# Each job reads its own secret, `DEPLOY_STAGING_IDENTITY_PEM` on the
+# `playground` environment and `DEPLOY_PRODUCTION_IDENTITY_PEM` on `production`,
+# so which key a job holds is obvious from the job and not from environment
+# scoping alone.
 #
 # Production is deliberately not wired up by default. The production key is not
 # held here; production is deployed from an operator machine with
@@ -77,7 +79,7 @@ jobs:
           DEPLOY_WALLET: ${{ inputs.wallet }}
           DEPLOY_MARKETING: ${{ inputs.marketing }}
           DEPLOY_DOCS: ${{ inputs.docs }}
-          DEPLOY_IDENTITY_PEM: ${{ secrets.DEPLOY_IDENTITY_PEM }}
+          IDENTITY_PEM: ${{ secrets.DEPLOY_STAGING_IDENTITY_PEM }}
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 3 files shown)


.github/workflows/release-deploy-frontend.yaml
```diff
@@ -0,0 +1,137 @@
...
+          persist-credentials: false
+      - name: 'Install icx-asset'
+        run: cargo install --locked icx-asset --version 0.29.0
+      - name: 'Deploy to production'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          DEPLOY_WALLET: ${{ inputs.wallet }}
+          DEPLOY_MARKETING: ${{ inputs.marketing }}
+          DEPLOY_DOCS: ${{ inputs.docs }}
+          DEPLOY_IDENTITY_PEM: ${{ secrets.DEPLOY_IDENTITY_PEM }}
```
aterga: Can we rename this to 
```suggestion
          DEPLOY_STAGING_IDENTITY_PEM: ${{ secrets.DEPLOY_STAGING_IDENTITY_PEM }}
```

(maybe in a follow-up PR)

.github/workflows/release-deploy-frontend.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+          for app in "${apps[@]}"; do
+            echo "==> deploying $app to playground"
+            ./scripts/deploy-app --app "$app" --env playground --pem "$pem" --yes
+          done
+
+  production:
+    name: 'production'
+    needs: playground
+    if: ${{ inputs.promote_to_production }}
```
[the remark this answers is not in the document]
MRmarioruci: Right about the type mismatch, though the fix that landed went the wrong way. A `workflow_dispatch` `type: boolean` input is a real boolean in the expression context, so `inputs.promote_to_production == 'true'` coerces both sides to numbers, `'true' becomes NaN, and the condition could never be satisfied. The production job was unreachable either way. Now just `if: ${{ inputs.promote_to_production }}`. The shell steps keep their string comparisons, which is correct there since env vars do stringify to "true"/"false".

scripts/deploy-app
```diff
@@ -0,0 +1,165 @@
...
+    *) usage; die "unknown argument: $1" ;;
+  esac
+done
+
+[[ -n "${BUILD_TARGET[$APP]:-}" ]] || { usage; die "--app must be one of: wallet marketing docs"; }
+[[ "$ENV" == "playground" || "$ENV" == "production" ]] || { usage; die "--env must be playground or production"; }
+[[ -n "$PEM" || -n "$OP_REF" ]] || { usage; die "a key is required: pass --pem or --op"; }
+[[ -z "$PEM" || -z "$OP_REF" ]] || die "pass only one of --pem / --op"
+[[ -z "$PEM" || -f "$PEM" ]] || die "pem file not found: $PEM"
+
```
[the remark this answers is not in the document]
MRmarioruci: Already mutually exclusive: `[[ -z "$BUILD" || -z "$TAG" ]] || die "pass only one of --build / --tag"`. Playground additionally rejects `--tag` outright, since the released artifact is a production build.

.github/workflows/release-deploy-frontend.yaml
MRmarioruci: Done in this PR rather than a follow-up, since nothing here is merged yet.

Split it per environment instead of one shared name, because your comment landed on the production job's line and naming that one `STAGING` would have been wrong: the playground job now reads `DEPLOY_STAGING_IDENTITY_PEM` and the production job reads `DEPLOY_PRODUCTION_IDENTITY_PEM`. Same split on the backend workflow (`BACKEND_STAGING_IDENTITY_PEM` / `BACKEND_PRODUCTION_IDENTITY_PEM`). Which key a job holds is now visible in the job instead of implied by environment scoping.

One thing worth your opinion: the environment is still called `playground`, matching `canister_ids.json` and `deploy-app --env playground`, so the secret says staging while the environment says playground. Happy to make it `DEPLOY_PLAYGROUND_IDENTITY_PEM` for consistency if you prefer, or rename the environment separately.

Also relevant since you are looking at this line: production is no longer meant to run from CI at all. The production key stays out of the repo and production is deployed with `scripts/deploy-app --op` from an operator machine. The production job is kept wired but inert.

scripts/deploy-app
```diff
@@ -0,0 +1,165 @@
...
+  echo "==> downloading ${TARBALL[$APP]} from $TAG"
+  gh release download "$TAG" --repo "$REPO" --pattern "${TARBALL[$APP]}" --dir "$DIR" --clobber
+else
+  echo "==> building for $ENV"
+  BUILD_MODE="$ENV" ./scripts/docker-build.sh "${BUILD_TARGET[$APP]}"
+fi
+
+# 2. Unpack.
+echo "==> unpacking"
+rm -rf "$DIR/dist" && mkdir -p "$DIR/dist"
```
aterga: Using a temporary directory could solve this problem in a non-distractive way

aterga (approved):

scripts/deploy-app
MRmarioruci: done

.github/workflows/release-deploy-frontend.yaml
MRmarioruci: Correction on my earlier reply: I have reverted the `STAGING` half of this.

Checking properly, `staging` is a separate network in `dfx.json`, offered by `scripts/deploy.sh --staging` and listed in HACKING.md, but it has no entries in `canister_ids.json`, so it is declared and empty. The environment with real canisters and a live site is `playground`. Naming the key `STAGING` therefore added a third word for the environment while pointing at a different, dead one.

So it is now `DEPLOY_PLAYGROUND_IDENTITY_PEM` / `DEPLOY_PRODUCTION_IDENTITY_PEM`, and the same on the backend workflow, tracking the network names the repo already uses.

The part of your suggestion I kept is the split itself, which was the right call: one shared name meant you had to know the environment scoping to know which key a job held.

If you did mean the `staging` network specifically, say so and I will wire it up properly instead, but it would need canister ids first.

PR #658, opened 2026-09-01
docs(release): add release process guide

MRmarioruci:
Adds `RELEASE.md`, an operator guide to the release process.

It opens with the component table, since that is where the confusion lives: what each project is, which pieces are singletons and which are per-org pairs, what an `-api` crate is versus the canister that runs, and that `dfx-orbit` and `orbit-cli` are different tools. Each row carries the current name beside a proposed one, so a rename can be discussed separately from this change.

Then the three phases: which boxes to tick to cut a release, how publish fires on merge with nothing to click, and how deploy splits into playground workflows and production commands, with the reason for that split and the exact commands. It ends with the one-time environment setup, including why the identity that deploys today must not be reused for CI.

Also corrects the Phase 1 section, which described a comma-separated `projects` field after that input became one checkbox per project.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,12 +10,14 @@ There is no single "Orbit release". There are independently versioned projects, 
 
 The current name is fixed. It is the git tag (`@orbit/{name}-v{version}`), the artifact name, and the build target, so renaming it is a real change, not a label. The proposed name is the clearer one we would move to in a separate PR if we decide to.
 
+Prose here says station, not wallet. Identifiers still say wallet, because that is what they literally are: the `wallet-dapp` project, the `wallet` checkbox on the deploy form, `--app wallet`, the `app_wallet` key in `canister_ids.json`. Changing those is the rename PR, not a wording fix.
+
 | Current name | Proposed name | Type | Role and where it deploys | Ships alone |
 | --- | --- | --- | --- | --- |
-| `wallet-dapp` | `station-frontend` | Frontend | The wallet UI users log into. Asset tarball to the wallet canister (`5fu67`, app.orbit.global). | Yes |
+| `wallet-dapp` | `station-frontend` | Frontend | The station UI users log into. Asset tarball to its asset canister (`5fu67`, app.orbit.global). | Yes |
 | `marketing-dapp` | `landing-page` | Frontend | The public marketing site. Asset tarball to the marketing canister (orbit.global). | Yes |
 | `docs-portal` | `docs-portal` | Frontend | The documentation site. Asset tarball to the docs canister (docs.orbit.global). | Yes |
-| `station` | `station` | Backend | Per-wallet backend, one deployed per org. Wasm to the control-panel registry, stations self-upgrade. | Yes |
+| `station` | `station` | Backend | The backend, one instance deployed per org. Wasm to the control-panel registry, stations self-upgrade. | Yes |
 | `upgrader` | `upgrader` | Backend | Per-station helper that performs safe upgrades, paired one-to-one with a station. Wasm to the registry. | Yes |
 | `control-panel` | `control-panel` | Backend | The single global registry and directory. Deploys stations. Wasm, deployed as the control-panel canister. | Yes |
 | `dfx-orbit` | `orbit-cli`, but see below | CLI | The CLI we ship to users. Git tag and GitHub release, users install it themselves. | Yes |
@@ -23,7 +25,7 @@ The current name is fixed. It is the git tag (`@orbit/{name}-v{version}`), the a
 
 Things worth stating plainly, because they are the usual source of confusion:
 
-* The three frontends are separate projects with separate versions and separate asset canisters. They are not one "wallet dapp" bundle.
+* The three frontends are separate projects with separate versions and separate asset canisters. They are not a single bundle.
 * Only the Control Panel is a singleton. There is one global instance. Station and Upgrader are multi-instance: the Control Panel deploys a fresh Station per org, and each Station comes paired with its own Upgrader. Station and Upgrader deploy and control each other, which is what makes a station upgrade safe.
 * The `station-api` / `upgrader-api` / `control-panel-api` crates are the contract, just the Candid interface and shared types. The bare name (`station`) is the canister that runs; the `-api` crate compiles to no canister. Bumping an api crate cascades a bump into whatever depends on it, which is why one small change can move several version numbers at once.
 * `dfx-orbit` and `orbit-cli` are not the same tool today. `dfx-orbit` is the CLI we ship to users. `orbit-cli` is our internal tool that drives the release (`release prepare`, `release publish`, `registry publish`). Only `dfx-orbit` is a release target. Which is why the rename in the table above has a catch: giving the user-facing CLI the name `orbit-cli` means renaming the internal one in the same change, or the name refers to two different tools. Every other row in that table is a straight rename; this one is the exception.
@@ -57,9 +59,9 @@ Both halves run the same two scripts, `scripts/deploy-app` and `scripts/deploy-b
 
 ### Frontends to playground
 
-Actions tab, run **Deploy frontend**. Tick **wallet** and run it. It builds the wallet for playground and uploads it to `bxkhk-6yaaa-aaaal-ai6va-cai`. Go test at https://playground.orbitwallet.io.
+Actions tab, run **Deploy frontend**. Tick **wallet** and run it. It builds the station frontend for playground and uploads it to `bxkhk-6yaaa-aaaal-ai6va-cai`. Go test at https://playground.orbitwallet.io.
 
-Only the wallet has a playground canister. `marketing-dapp` and `docs-portal` exist on production only, so ticking them here fails on purpose, with a message telling you to deploy them to production instead. The backends are unaffected: all three targets work on playground, since the registry publish path needs only the control-panel and the wasm chunk store, both of which exist there.
+Only the station frontend has a playground canister. `marketing-dapp` and `docs-portal` exist on production only, so ticking them here fails on purpose, with a message telling you to deploy them to production instead. The backends are unaffected: all three targets work on playground, since the registry publish path needs only the control-panel and the wasm chunk store, both of which exist there.
 
 Leave **promote_to_production** unchecked. It is off by default and only does anything if a production key has been put on the environment, which is not how this is set up.
 
@@ -68,7 +70,7 @@ Leave **promote_to_production** unchecked. It is off by default and only does an
 Actions tab, run **Deploy backend**. Tick station, upgrader, or control-panel and run it. Same as above: leave promote unchecked. The two backend types deploy differently:
 
 * **station / upgrader**: `orbit-cli registry publish` loads the wasm into the control-panel registry. This only makes the version available. Each production station still upgrades itself, or its upgrader, only when that station's own admins submit and approve the request (finance being 2-of-N). Publishing station also publishes upgrader, its dependency.
-* **control-panel**: a direct `dfx canister install --mode upgrade` of the one canister. There is no registry and no per-wallet buffer, so nothing stands between this and every wallet. Handle with care.
+* **control-panel**: a direct `dfx canister install --mode upgrade` of the one canister. There is no registry and no per-station buffer, so nothing stands between this and every station. Handle with care.
 
 ### Production
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 
 A release has three phases: cut, publish, deploy. Cutting and publishing are one flow that ends in a GitHub release with the built artifact attached. Deploying is what puts that artifact on the live canisters. All of it runs from the Actions tab. This is the guide to what to click.
 
-Most of this used to be manual, run from a laptop with the production key. It is now three workflows: **Cut release**, **Deploy frontend**, **Deploy backend**. Publishing was already automated and has not changed.
+Most of this used to be manual. It is now three workflows: **Cut release**, **Deploy frontend**, **Deploy backend**. Publishing was already automated and has not changed.
 
 ## Components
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,86 @@
...
+
+| Current name | Proposed name | Type | Role and where it deploys | Ships alone |
+| --- | --- | --- | --- | --- |
+| `wallet-dapp` | `wallet-app` | Frontend | The wallet UI users log into. Asset tarball to the wallet canister (`5fu67`, app.orbit.global). | Yes |
+| `marketing-dapp` | `landing-page` | Frontend | The public marketing site. Asset tarball to the marketing canister (orbit.global). | Yes |
+| `docs-portal` | `docs-site` | Frontend | The documentation site. Asset tarball to the docs canister (docs.orbit.global). | Yes |
+| `station` | `station` | Backend | Per-wallet backend, one deployed per org. Wasm to the control-panel registry, stations self-upgrade. | Yes |
+| `upgrader` | `upgrader` | Backend | Per-station helper that performs safe upgrades, paired one-to-one with a station. Wasm to the registry. | Yes |
+| `control-panel` | `control-panel` | Backend | The single global registry and directory. Deploys stations. Wasm, deployed as the control-panel canister. | Yes |
+| `dfx-orbit` | `orbit-cli` | CLI | The CLI we ship to users. Git tag and GitHub release, users install it themselves. | Yes |
```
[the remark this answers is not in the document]
MRmarioruci: Correct, the doc contradicted itself. Kept the proposal but made the catch explicit: taking `orbit-cli` for the user-facing CLI means renaming the internal one in the same change, which makes it the one row in that table that is not a straight rename.

RELEASE.md
```diff
@@ -0,0 +1,86 @@
...
+* The `station-api` / `upgrader-api` / `control-panel-api` crates are the contract, just the Candid interface and shared types. The bare name (`station`) is the canister that runs; the `-api` crate compiles to no canister. Bumping an api crate cascades a bump into whatever depends on it, which is why one small change can move several version numbers at once.
+* `dfx-orbit` and `orbit-cli` are not the same tool. `dfx-orbit` is the CLI we ship to users. `orbit-cli` is our internal tool that drives the release (`release prepare`, `release publish`, `registry publish`). Only `dfx-orbit` is a release target.
+
+## Phase 1: cut a release
+
+Actions tab, run the **Cut release** workflow. The form:
+
+* **projects**: comma-separated project names, for example `wallet-dapp,station`. Empty releases everything that changed since the last release. This is where you pick the whole batch or a subset.
```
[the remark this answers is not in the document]
MRmarioruci: Fixed. That bullet predated the input becoming one checkbox per project and never got updated.

RELEASE.md
```diff
@@ -0,0 +1,86 @@
...
+* **version_specifier**: `auto` lets the conventional commits decide the bump. Override with `patch` / `minor` / `major` if you need to.
+* **pre_release**: `none`, or `alpha` / `beta` / `rc` for a pre-release.
```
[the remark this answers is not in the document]
MRmarioruci: Fixed. The specifier bullet now lists the `prepatch` / `preminor` / `premajor` / `prerelease` modes, and the `pre_release` bullet says it is only valid with `auto` or `prerelease`, which matches the guard in the workflow.

RELEASE.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Components
+
+There is no single "Orbit release". There are independently versioned projects, and nx bumps each one on its own from the conventional commits that touched it. Six produce a deployable artifact. The rest are internal crates that only exist to cascade a version bump.
+
+The current name is fixed. It is the git tag (`@orbit/{name}-v{version}`), the artifact name, and the build target, so renaming it is a real change, not a label. The proposed name is the clearer one we would move to in a separate PR if we decide to.
+
+| Current name | Proposed name | Type | Role and where it deploys | Ships alone |
+| --- | --- | --- | --- | --- |
+| `wallet-dapp` | `wallet-app` | Frontend | The wallet UI users log into. Asset tarball to the wallet canister (`5fu67`, app.orbit.global). | Yes |
```
aterga:
```suggestion
| `wallet-dapp` | `station-frontend` | Frontend | The wallet UI users log into. Asset tarball to the wallet canister (`5fu67`, app.orbit.global). | Yes |
```

RELEASE.md
```diff
@@ -0,0 +1,99 @@
...
+| `marketing-dapp` | `landing-page` | Frontend | The public marketing site. Asset tarball to the marketing canister (orbit.global). | Yes |
+| `docs-portal` | `docs-site` | Frontend | The documentation site. Asset tarball to the docs canister (docs.orbit.global). | Yes |
```
aterga:
```suggestion
| `docs-portal` | `docs-portal` | Frontend | The documentation site. Asset tarball to the docs canister (docs.orbit.global). | Yes |
```

MRmarioruci:
Done, and swept the whole stack rather than just this PR, since the same wording was in the workflow files too.

In `RELEASE.md`: the station row no longer says "Per-wallet backend", the frontend row describes the station UI rather than the wallet UI, "no per-wallet buffer ... every wallet" is now station, and the "wallet dapp bundle" phrasing is gone.

Upstream, because these render in the Actions form and read worse than the doc did: dfinity/orbit#655 'Release wallet-dapp (wallet app)' is now '(the station frontend)', dfinity/orbit#656 'Deploy the wallet app' is now 'Deploy the station frontend' plus a comment, and dfinity/orbit#657 loses "per-wallet buffer" and "every wallet".

What still says wallet is only the literal identifiers: the `wallet-dapp` project name and tag, the `wallet` checkbox on the deploy form, `--app wallet`, and the `app_wallet` key in `canister_ids.json`. Those are names rather than prose, so changing them is the rename PR (where `wallet-dapp` becomes `station-frontend`), not a wording fix. The doc now says that explicitly so it does not read as something we missed.

No literal wallet page with a token balance view appears in these files, so that exception did not come up.

RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+* Only the Control Panel is a singleton. There is one global instance. Station and Upgrader are multi-instance: the Control Panel deploys a fresh Station per org, and each Station comes paired with its own Upgrader. Station and Upgrader deploy and control each other, which is what makes a station upgrade safe.
+* The `station-api` / `upgrader-api` / `control-panel-api` crates are the contract, just the Candid interface and shared types. The bare name (`station`) is the canister that runs; the `-api` crate compiles to no canister. Bumping an api crate cascades a bump into whatever depends on it, which is why one small change can move several version numbers at once.
+* `dfx-orbit` and `orbit-cli` are not the same tool today. `dfx-orbit` is the CLI we ship to users. `orbit-cli` is our internal tool that drives the release (`release prepare`, `release publish`, `registry publish`). Only `dfx-orbit` is a release target. So the rename in the table has a catch: giving the user-facing CLI the name `orbit-cli` means renaming the internal one in the same change, or the name means two different tools. Every other row is a straight rename.
```
aterga: Let's rename the internal tool into `orbit-release-cli`

RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+| `upgrader` | `upgrader` | Backend | Per-station helper that performs safe upgrades, paired one-to-one with a station. Wasm to the registry. | Yes |
+| `control-panel` | `control-panel` | Backend | The single global registry and directory. Deploys stations. Wasm, deployed as the control-panel canister. | Yes |
+| `dfx-orbit` | `orbit-cli`, but see below | CLI | The CLI we ship to users. Git tag and GitHub release, users install it themselves. | Yes |
+| `*-api`, `orbit-essentials` | (unchanged) | Crates | Shared Candid and types. No artifact, deploys nowhere. | Bumped automatically |
+
+The usual points of confusion:
+
+* The three frontends are separate projects with separate versions and separate asset canisters. They are not a single bundle.
+* Only the Control Panel is a singleton. There is one global instance. Station and Upgrader are multi-instance: the Control Panel deploys a fresh Station per org, and each Station comes paired with its own Upgrader. Station and Upgrader deploy and control each other, which is what makes a station upgrade safe.
```
aterga:
```suggestion
* Apart from the three frontends, only the Control Panel is a singleton. There is one global instance. Station and Upgrader are multi-instance: the Control Panel deploys a fresh Station per org, and each Station comes paired with its own Upgrader. Station and Upgrader deploy and control each other, which is what makes a station upgrade safe.
```

RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+* The `station-api` / `upgrader-api` / `control-panel-api` crates are the contract, just the Candid interface and shared types. The bare name (`station`) is the canister that runs; the `-api` crate compiles to no canister. Bumping an api crate cascades a bump into whatever depends on it, which is why one small change can move several version numbers at once.
+* `dfx-orbit` and `orbit-cli` are not the same tool today. `dfx-orbit` is the CLI we ship to users. `orbit-cli` is our internal tool that drives the release (`release prepare`, `release publish`, `registry publish`). Only `dfx-orbit` is a release target. So the rename in the table has a catch: giving the user-facing CLI the name `orbit-cli` means renaming the internal one in the same change, or the name means two different tools. Every other row is a straight rename.
+
+## Phase 1: cut a release
+
+Actions tab, run the **Cut release** workflow. The form:
+
+* **one checkbox per project**: tick `wallet-dapp`, `station`, and so on. Tick nothing and it releases everything that changed since the last release, which nx works out from the commits. This is where you pick the whole batch or a subset.
```
aterga:
```suggestion
* **one checkbox per project**: tick `station-frontend`, `station`, and so on. Tick nothing and it releases everything that changed since the last release, which nx works out from the commits. This is where you pick the whole batch or a subset.
```

aterga: Let's use the new names consistently throughout.

RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+* The `station-api` / `upgrader-api` / `control-panel-api` crates are the contract, just the Candid interface and shared types. The bare name (`station`) is the canister that runs; the `-api` crate compiles to no canister. Bumping an api crate cascades a bump into whatever depends on it, which is why one small change can move several version numbers at once.
+* `dfx-orbit` and `orbit-cli` are not the same tool today. `dfx-orbit` is the CLI we ship to users. `orbit-cli` is our internal tool that drives the release (`release prepare`, `release publish`, `registry publish`). Only `dfx-orbit` is a release target. So the rename in the table has a catch: giving the user-facing CLI the name `orbit-cli` means renaming the internal one in the same change, or the name means two different tools. Every other row is a straight rename.
+
+## Phase 1: cut a release
+
+Actions tab, run the **Cut release** workflow. The form:
```
aterga, line 38: I don't see this there yet, is that expected? 

Image

MRmarioruci: yes. You will only see it once the stack is merged

RELEASE.md
MRmarioruci: Well this refers to the checkbox that the user will see and I dont want to change those. Cause it's the project id. If i change it on the checkbox/workflow this means I need to change the project ids etc which is more work and will result in more confusion on the releases page (historical releases + new ones). I can do it.. I am not sure if this is worth it or not at least for now.

RELEASE.md
MRmarioruci: Done. The internal tool now has its own row proposing `orbit-release-cli`, and the `dfx-orbit` row is a straight rename to `orbit-cli` like the rest of the table.

Worth noting that this one is cheap to actually do, unlike the others. The internal CLI has no entry in `.release.json`, so it has never been tagged or published and there is no release history to keep consistent. The change is `cli/package.json` (name and bin), the `prepare-cli` script in the root `package.json`, and the call sites in `release-cut.yaml`, `release.yaml`, the deploy-backend action and `scripts/deploy-backend`. Happy to do it as a separate PR whenever you want it, independently of the project renames.

RELEASE.md
MRmarioruci: Expected. GitHub only lists `workflow_dispatch` workflows that exist on the default branch, and `release-cut.yaml` is still only on this stack. Confirmed with `gh workflow list`, which shows Code benchmarks, Deterministic builds, Release, Code tests, Code validation, Copilot and Dependabot, but none of the three new ones.

The form appears once dfinity/orbit#655 merges to `main`. Until then the workflow cannot be run at all, from any branch, which is also why none of this has been exercised end to end yet.

RELEASE.md
MRmarioruci: Agreed on the goal, but I did not take the suggestion literally, because the form field is genuinely called `wallet_dapp` today. If the doc says to tick `station-frontend`, an operator reading it next week looks at the Actions form and finds no such checkbox.

So I stated the convention instead and made the doc stop depending on the name: backticked names are the current ones, matching the repo and the workflow forms, and the table maps each to its proposed name. The Phase 1 bullet now says "one checkbox per project, named as in the table above" rather than naming one, so it stays correct through the rename. Where the doc has to name a form field, it says both, as in "tick the station frontend, which the form still calls **wallet**".

Prose is all station now. What still says wallet is only literal identifiers, and the paragraph above the table explains why they are expensive to change: the project name is the release tag prefix, so renaming splits the tag series and the scripts resolving `latest` would need to match both names.

If you would rather the doc describe the post-rename world and we accept it being ahead of the code, say so and I will switch it. It would be better to land the rename PR first, and I am happy to write it.

RELEASE.md
MRmarioruci: Done: dfinity/orbit#662, stacked on this PR.

It sits on top of this one rather than on `main` because `release-cut.yaml`, `scripts/deploy-backend` and `RELEASE.md` only exist in this stack. Landing the rename on `main` first would leave those files calling a binary that no longer exists the moment they merge.

Verified `pnpm install` runs the renamed `prepare-cli` end to end and the built CLI reports `orbit-release-cli` with all three subcommands intact. `git grep orbit-cli` now returns one hit, the proposed name for `dfx-orbit` in the table.

MRmarioruci: Follow-up: dfinity/orbit#662 is now based on `main` instead of on this PR, so it no longer sits in front of the stack. #655 through #658 can merge on their own.

To make the order genuinely not matter, #662 keeps `orbit-cli` as a second `bin` entry alongside `orbit-release-cli`. The three call sites this stack adds (`release-cut.yaml`, the deploy-backend action, `scripts/deploy-backend`) still say `orbit-cli` and do not exist on `main`, so they cannot be renamed there. The alias means neither PR has to wait for the other, and nothing ends up calling a binary that is not present.

Once the stack lands I will open a small follow-up that renames those three and removes the alias. `RELEASE.md` in this PR still describes `orbit-release-cli` as the proposed name, which is accurate until #662 merges.

PR #661, opened 2026-09-04
docs(readme): state what components do and split shipped from planned

MRmarioruci:
Rewrites `README.md` in direct language: say what each component does rather than what it is responsible for, and separate what ships from what is planned.

The component list now names the actual behavior. The Control Panel deploys a Station on request, stores the Station and Upgrader wasms that stations upgrade to, and lists the stations a user belongs to. The Station holds the assets, users, permissions and approval policies, and carries out an operation only once that operation's policy is satisfied. The Upgrader is paired one-to-one with a Station, installs new Station wasms, and can restore the Station when an upgrade fails. Previously three of the four bullets said "is responsible for" or "facilitates common operations" without saying which.

The Target Architecture section listed the CDK, SDK and extensions marketplace in the same breath as shipped components, with the distinction left to the diagram's highlighting. It now says which boxes ship today and which are planned. Multi-chain support likewise read as current and is now marked as not built yet.

Two factual corrections. `orbit.global` is the landing page and the wallet runs at `app.orbit.global`, but the intro link and the applications section both pointed at `orbit.global` for the app. And the tests badge linked to `./actions/workflows/tests.yaml`, which GitHub resolves inside the repo tree and 404s.

Unrelated to the release workflow stack (#655 through #658), so it is branched off `main` and can merge independently. No content was removed except one closing line that repeated the point above it, and the diagrams, commands, badges and links are otherwise untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,26 +1,30 @@
-> This project is **still in Alpha**. We are happy to answer questions if they are raised as issues in this github repo.
+> This project is **still in Alpha**. We are happy to answer questions if they are raised as issues in this GitHub repo.
 
 [![Internet Computer portal](https://img.shields.io/badge/InternetComputer-grey?logo=internet%20computer&style=for-the-badge)](https://internetcomputer.org)
 [![GitHub license](https://img.shields.io/badge/license-Apache%202.0-blue.svg?logo=apache&style=for-the-badge)](LICENSE)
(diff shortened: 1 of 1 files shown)


MRmarioruci:
Wrong file, this was meant for RELEASE.md. Closing.

